PR #144, opened 2014-12-17
Set billboard opacity

gberaudo:
Handle opacity property in the converter.

### DIFF
--- a/Cesium.externs.js
+++ b/Cesium.externs.js
@@ -112,6 +112,7 @@ Cesium.BillboardCollection = function() {};
 /**
  * @typedef {{
  *   image: (string|HTMLCanvasElement|Image),
+ *   color: (Cesium.Color|undefined),
  *   verticalOrigin: (Cesium.VerticalOrigin|undefined),
  *   horizontalOrigin: (Cesium.HorizontalOrigin|undefined),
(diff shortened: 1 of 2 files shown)


ahocevar:
Looks good. Please merge.

schmidtk:
This works if the opacity is set at the style level, but it may be set at the layer level. Are there plans to handle that case?

gberaudo:
Unfortunately, this OL3 mechanism does not have equivalent in Cesium.
A good solution would avoid destroying and recreating all primitives.
It would require Ol3 to propagate layer changes in a finer way.

ahocevar:
I'll take a look. Maybe there is a way.

ahocevar:
@schmidtk As far as I can tell, this change also works for an opacity set on the layer level. The style that is queried for the opacity still comes from `olcs.core.computePlainStyle`, which takes both feature and layer styles into account and executes the style function for each feature. Can you confirm?

ahocevar:
I see now what you mean @schmidtk: an opacity set by `ol.layer.Base#setOpacity`. I'll see if I can come up with something.

ahocevar:
@schmidtk I came up with #166 for layer opacity support.

PR #147, opened 2015-01-08
Allow working with custom synchronizers

ahocevar:
This change allows developers to configure a `createSymbolizers` function for OLCesium, instead of relying on the default set of synchronizers.

### DIFF
--- a/externs/olcsx.js
+++ b/externs/olcsx.js
@@ -5,6 +5,45 @@
  */
 var olcsx;
 
+
+/**
+ * @typedef {{
+ *   map: (!ol.Map),
+ *   target: (Element|string|undefined),
+ *   createSynchronizers: ((function(!ol.Map, !Cesium.Scene): Array.<olcs.AbstractSynchronizer>)|undefined)
+ * }}
+ * @api
+ */
+olcsx.OLCesiumOptions;
+
+
+/**
+ * The OpenLayers map we want to show on a Cesium scene.
+ * @type {!ol.Map}
+ * @api
+ */
+olcsx.OLCesiumOptions.prototype.map;
+
+
+/**
+ * Target element for the Cesium scene.
+ * @type {Element|string|undefined}
+ * @api
+ */
+olcsx.OLCesiumOptions.prototype.target;
+
+
+/**
+ * Callback function which will be called by the {@link olcs.OLCesium}
+ * constructor to create custom synchronizers. Receives an `ol.Map` and a
+ * `Cesium.Scene` as arguments, and needs to return an array of
+ * {@link olcs.AbstractSynchronizer}.
+ * @type {undefined|function(!ol.Map, !Cesium.Scene): Array.<olcs.AbstractSynchronizer>}
+ * @api
+ */
+olcsx.OLCesiumOptions.prototype.createSynchronizers;
+
+
 /**
  * Core namespace.
  * @type {Object}
--- a/src/ol3cesium.js
+++ b/src/ol3cesium.js
@@ -11,12 +11,24 @@ goog.require('olcs.VectorSynchronizer');
 
 
 /**
- * @param {!ol.Map} map
+ * @param {!(olcsx.OLCesiumOptions|ol.Map)} options Options.
  * @param {Element|string=} opt_target Target element for the Cesium container.
+ *   Use the `target` option in `options` instead.
  * @constructor
  * @api
  */
-olcs.OLCesium = function(map, opt_target) {
+olcs.OLCesium = function(options, opt_target) {
+
+  var map, createSynchronizers;
+  var target = opt_target;
+  if (options instanceof ol.Map) {
+    map = options;
+  } else {
+    map = options.map;
+    target = options.target;
+    createSynchronizers = options.createSynchronizers;
+  }
+
   /**
    * @type {!ol.Map}
    * @private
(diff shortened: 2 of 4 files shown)


externs/olcsx.js
```diff
@@ -5,6 +5,45 @@
  */
 var olcsx;
 
+
+/**
+ * @typedef {{
+ *   map: (!ol.Map),
+ *   target: (Element|string|undefined),
+ *   createSynchronizers: (undefined|function(!ol.Map, !Cesium.Scene): Array.<olcs.AbstractSynchronizer>)
```
gberaudo: I agree in the case of optional functions that putting undefined first is more readable.
However, it is not the way it is done in ol3. Do we want to add an exception to the coding style?

ahocevar: In the case of functions (which we have here) the same is done in ol3. Otherwise you need more braces.

src/ol3cesium.js
```diff
@@ -11,12 +11,26 @@ goog.require('olcs.VectorSynchronizer');
...
 /**
- * @param {!ol.Map} map
+ * @param {!(ol.Map|olcsx.OLCesiumOptions)} arg1 `ol.Map` or options
  * @param {Element|string=} opt_target Target element for the Cesium container.
+ *   Overrides a target set in options.
  * @constructor
  * @api
  */
-olcs.OLCesium = function(map, opt_target) {
+olcs.OLCesium = function(arg1, opt_target) {
```
gberaudo: The old behaviour of taking a map and an optional target must be deprecated and will be removed in the near future. Additionally, a comment explaining how to switch to the new behaviour must be added.
We will later add a CHANGES.md file like in Cesium `https://github.com/AnalyticalGraphicsInc/cesium/blob/master/CHANGES.md`.

ahocevar: Allowing both ways does not add significant overhead, so I'm inclined to not deprecate the current signature. It also adds convenience if you are not interested in custom synchronizers.

externs/olcsx.js
gberaudo: I checked the ol3 code with `git grep 'undefined|function'` and found no occurrence.

gberaudo:
Thanks @ahocevar, for the PR.
It looks good; I wrote two comments, one about deprecating the old constructor and explaining how to switch to the new behaviour.

externs/olcsx.js
ahocevar: For example [here](https://github.com/openlayers/ol3/blob/master/externs/olx.js#L6117) and [here](https://github.com/openlayers/ol3/blob/master/externs/olx.js#L6131).

gberaudo: You checked the only 2 cases where `undefined` is at the beginning in ol3.
On the contrary, putting `undefined` at the end is done 24 times.
Checked with `git grep -i 'undefined|.*function'` and `git grep -i '.*Function|undefined'`

src/ol3cesium.js
gberaudo: The ol3cesium constructor will definitively allow more and more customisations making it necessary to use the new behaviour.
It was an error to design this constructor this way and we have now the responsibility to fix it before more and more users start relying on it.

It is straightforward to switch to the new way, thus we can have a short deprecation period of 1 'minor' release: release 1.1.0 will have the new behaviour but still accept the old behaviour with deprecation notice stating that the old behaviour will be removed in the 1.2.0 release. At that time, the CHANGES.md will be updated with a note about this change.

ahocevar: [Semantic versioning](http://semver.org) works differently. I'm going to add a note, however.

externs/olcsx.js
ahocevar: I'm going to change it so it suits your taste.

src/ol3cesium.js
gberaudo: I do not think the way versioning is handled in ol3 is appropriate for
ol3cesium: ol3-cesium is very new compared to the openlayers project.

A better model for ol3-cesium is Cesium, which is a 2 year old project.

ahocevar: Semantic versioning is nothing specific to ol3, it is a de-facto standard
in the JavaScript library world. But we should continue this discussion in
a dedicated place, and not in this issue.

tschaub: Keep in mind that if you are really attached to a particular version number you can always go with [Sentimental Versioning](http://sentimentalversioning.org/) instead.  That would let you pick your favorite version numbers and break things without feeling bad.

gberaudo:
@ahocevar, you have an incorrect manner to take into account reviews.

Creating a PR and merging it whatever the review comments is not acceptable.
I remind you it is the second time you do it. See #119 which reverts #118.

Please fix your code with a `@deprecated` tag on the constructor stating the old behaviour is deprecated and the new behaviour must be used. Additionally, add a console log in the constructor body, which will be triggered when the developer uses the old behaviour, and inform the developer about the deprecation and migration path.

The time when we will remove the old behaviour may be discussed later, though it would be better to have it now.

I am forced to revert the merge so that you can improve your code by addressing the review comments.

bartvde:
@gberaudo any chance that instead of reverting someone's work you can add a commit / PR that adds the deprecation warning? It's also important to remain constructive IMHO. It seems you feel more strongly about the deprecation path. It seems to be simply something for which there is no agreement between the author of the PR and the reviewer, which can happen.

Issue #149, opened 2015-01-09
How to handle breaking API changes


ahocevar:
With #147, a second signature for the `olcs.OLCesium` constructor was introduced. @gberaudo suggested to mark the old signature with a `@deprecated` annotation and a console message that reminds developers to change their code to use the new signature.

This ticket is about deciding how we handle breaking API changes. My suggestion was to use [Semantic Versioning](http://semver.org/). @tschaub suggested that we could also use [Sentimental Versioning](http://sentimentalversioning.org) instead. That would let us pick our favorite version numbers and break things without feeling bad. What @gberaudo suggests (add a deprecation warning and break things two point releases after the warning was added) would fall into that category.

What do others think?

elemoine:
> @tschaub suggested that we could also use Sentimental Versioning instead

which sounded like a joke to me…

Anyway, my opinion is that ol3-cesium is very young, and it's too early to favor background compatibility over a consistent API etc. I know v1.0.0 is out already, but I'd rather still allow us to break compatibility when necessary, and write changelogs for users to know how to migrate their applications.

bartvde:
+1 @elemoine

ahocevar:
Ok, I think this is in line with ol3, where we allow changes to experimental parts of the API also between minor versions. Without deprecation annotations/warnings, but with the changes described in the release changelog. So +1 from me too.

Ok if I make this breaking change on top of #150 and create a release notes draft in https://github.com/openlayers/ol3-cesium/releases/? @gberaudo?

gberaudo:
Before releasing Cesium 1.0, they also pushed breaking changes without
prior notice. Given the youth of the ol3-cesium project, I agree it is
reasonable to try the same approach.


When the design is proved and the API more stable, we will then be able
to switch to a more user-friendly approach where we announce future
changes and we wait a reasonable amount of time before removing deprecated
features.


Regarding how to communicate the library changes to the users, I would
like we add a CHANGES.md file, which will list breaking and non breaking
changes. This file will be updated in the PR which actually changes the
code, making it convenient for users to get details.
I have no opinion on how to organise the github release page yet.

Exemples from the Cesium project:
https://github.com/AnalyticalGraphicsInc/cesium/releases
https://github.com/AnalyticalGraphicsInc/cesium/blob/master/CHANGES.md

If there is disagreement in having a CHANGES.md file, let us discuss
about it now.


@ahocevar, if you find it convenient to use the PR #150 as a base please
[...]

PR #150, opened 2015-01-09
Restore reverted pull request #147

ahocevar:
My suggestion to resolve the discussion that was started with #147 about if or how to use deprecation warnings would be to first add the changes I made with #147, and then find an agreement about deprecation procedure (discussion in #149).

Thanks for any opinions and reviews.


bartvde:
I agree with this sequence, it makes sense to me to handle it this way. Once there is agreement on #149 there can be a follow-up PR.

ahocevar:
Closing in favour of #151.

PR #162, opened 2015-01-21
Handle features without geometry

### LINKED ISSUE #161: Unable to load features without geometries
While OL3 is perfectly capable of handling features which don't contain geometries, OL3-Cesium is not.

This may happen, for example, when loading GeoJSON data. Features without geometries [are perfectly valid GeoJSON](http://geojson.org/geojson-spec.html#feature-objects) data.

Consider the following valid GeoJSON structure:

<pre>
var geoJsonData = {
    "type": "FeatureCollection",
    "features": [
        {
            "type": "Feature",
            "id": "SOME_ID",
            "geometry": null,
            "properties": {
                "designator": "SOME_DESIGNATOR"
            }
        }
    ]
};
var features = new ol.format.GeoJSON().readFeatures(geoJsonData);
</pre>

When loading this into a vector source of a map with activated OL3-Cesium:

<pre>
someVectorSource.addFeatures(features);
</pre>

the following error message is thrown:

<pre>
Uncaught TypeError: Cannot read property 'getType' of null
</pre>

For reference, this is the workaround I'm currently using:

<pre>
var features = new ol.format.GeoJSON().readFeatures(geoJsonData);
features = features.filter(function(feature) {
    return feature.getGeometry() !== null;
});
someVectorSource.addFeatures(features);
</pre>

gberaudo:
Closes #161
Updated vectors example to include a feature without geometry.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1089,6 +1089,11 @@ goog.require('olcs.core.OlLayerPrimitive');
   olcs.core.olFeatureToCesium = function(feature, style, context, opt_geom) {
     var geom = opt_geom || feature.getGeometry();
     var proj = context.projection;
+    if (!geom) {
+      // Ol3 features may not have a geometry
+      // See http://geojson.org/geojson-spec.html#feature-objects
+      return null;
+    }
 
     var id = function(object) {
       object.olFeature = feature;
--- a/examples/vectors.js
+++ b/examples/vectors.js
@@ -106,7 +106,9 @@ var styles = {
 };
 
 var styleFunction = function(feature, resolution) {
-  return styles[feature.getGeometry().getType()];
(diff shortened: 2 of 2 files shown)


src/core.js
```diff
@@ -1089,6 +1089,11 @@ goog.require('olcs.core.OlLayerPrimitive');
   olcs.core.olFeatureToCesium = function(feature, style, context, opt_geom) {
     var geom = opt_geom || feature.getGeometry();
     var proj = context.projection;
+    if (!geom) {
```
fredj: Could be `goog.isNull(geom)`

fredj:
LGTM

gberaudo:
Thank you for the review and for the comment about a possible usage of `goog.isNull`.

PR #166, opened 2015-01-29
Support layer opacity for vector points

ahocevar:
See https://github.com/openlayers/ol3-cesium/pull/144#issuecomment-71260965. This adds support for layer opacity on vector points. Layer and style opacity are combined on the billboard level.


gberaudo:
@ahocevar, I will look at your code next week when I am back from FOSDEM.

schmidtk:
I'm guessing you'll need to define your OL opacity attributes in the Cesium externs to avoid making the compiler angry. Otherwise the approach looks sound to me.

src/core.js
```diff
@@ -789,9 +789,11 @@ goog.require('olcs.core.OlLayerPrimitive');
       var position = olcs.core.ol4326CoordinateToCesiumCartesian(center);
       var color;
       var opacity = imageStyle.getOpacity();
-      if (goog.isDef(opacity)) {
-        color = new Cesium.Color(1.0, 1.0, 1.0, opacity);
+      if (!goog.isDef(opacity)) {
+        opacity = 1;
       }
+      color = new Cesium.Color(1.0, 1.0, 1.0,
+          opacity * billboards.olLayerOpacity);
```
gberaudo: The `olLayerOpacity` will be undefined on first synchronization.
The correct way is to pass the layer to the synchronization function.

src/vectorsynchronizer.js
```diff
@@ -81,6 +81,22 @@ olcs.VectorSynchronizer.prototype.createSingleCounterpart = function(olLayer) {
     csPrimitives.show = olLayer.getVisible();
   });
 
+  var onOpacityChange = function() {
```
gberaudo: This code is unnecessary, on layer change, all the layer features are destroyed and created anew.
A finer mechanism in ol3-cesium would require the OL3 change event to be more detailed.

gberaudo:
@ahocevar, please see my comments.
Your PR only deals with Point geometries; do you plan to implement support for the other kinds of geometries to make it look consistent?

src/core.js
ahocevar: This is why `opacity` will be initialized to `1` above if it is undefined.

ahocevar:
@gberaudo I'll see if I can make this work for other geometry types as well. Points were a lower hanging fruit than the rest though.

src/core.js
gberaudo: Why `1`?

ahocevar: 1 means fully opaque, which is the default.

gberaudo: The layer opacity may be different from the default; that is why it is necessary to read the actual value.

ahocevar: Sorry, I think we had a misunderstanding. `billboards.olLayerOpacity` will never be undefined here. Only `opacity` can, and this is when no opacity is set on the style.

gberaudo: There are several code paths where `billboards.olLayerOpacity` will not be set.
See BillboardCollection instantiations.

src/vectorsynchronizer.js
ahocevar: It does not work without this code. Can you point me to the place where point features are destroyed and re-created upon layer change? Also, where are you envisioning more detailed ol3 change events? You can already listen for specific property changes on the layer, which is done here.

ahocevar:
@schmidtk:
 > I'm guessing you'll need to define your OL opacity attributes in the Cesium externs to avoid making the compiler angry.

Did you run `git submodule update` before building? The ol3 version currently used as submodule already has the layer opacity in the generated externs file.

src/vectorsynchronizer.js
gberaudo: You are right @ahocevar, we no longer listen for the global layer change event.
So all features are kept and it makes sense to try to update each object opacity.
The update code should be moved out of the synchronizer and placed in a core function.

src/vectorsynchronizer.js
```diff
@@ -81,6 +81,22 @@ olcs.VectorSynchronizer.prototype.createSingleCounterpart = function(olLayer) {
...
+    var bbs = csPrimitives.context.billboards;
+    var opacity = olLayer.getOpacity();
+    if (!goog.isDef(opacity)) {
+      opacity = 1;
+    }
+    bbs.olLayerOpacity = opacity;
+    var i, bb;
+    for (i = bbs.length - 1; i >= 0; --i) {
```
gberaudo: Why do you iterate in reverse order?

ahocevar: Starting the iterator with `bbs.length` saves me from having to store it in a separate variable. This is a small optimization I use frequently when the order of iteration does not matter.

src/vectorsynchronizer.js
```diff
@@ -81,6 +81,22 @@ olcs.VectorSynchronizer.prototype.createSingleCounterpart = function(olLayer) {
...
+      bb = bbs.get(i);
+      bb.color = new Cesium.Color(1.0, 1.0, 1.0, bb.olStyleOpacity * opacity);
```
gberaudo: You should keep the other  components of the colour.
There is a new function for this in Cesium, Color.fromAlpha.

ahocevar: The color of the billboard is not used for anything but setting the opacity. See https://github.com/openlayers/ol3-cesium/blob/3e9393c1817d28d8d6b17a437166833d49d0c1ac/src/core.js#L793. If that is a bad way to handle opacity, it should be changed there as well. Maybe something more suited for a follow-up pull request.

gberaudo: I know it is handled this way in the point converter. However, it may change in the future and there is no reason to hardcode a magic value here.

ahocevar: However, `Cesium.Color.fromAlpha` is not available yet in the Cesium version we currently use.

src/vectorsynchronizer.js
```diff
@@ -77,9 +77,10 @@ olcs.VectorSynchronizer.prototype.createSingleCounterpart = function(olLayer) {
   var csPrimitives = olcs.core.olVectorLayerToCesium(olLayer, view,
       featurePrimitiveMap);
 
-  olLayer.on('change:visible', function(e) {
-    csPrimitives.show = olLayer.getVisible();
+  olLayer.on(['change:visible', 'change:opacity'], function(e) {
+    olcs.core.updateCesiumPrimitives(olLayer, csPrimitives);
```
gberaudo: Changing visibility should not trigger the expensive modification of all opacities.

gberaudo:
@ahocevar, I feel it would be better to merge this PR in a dedicated branch of
the project until support for all other geometries are added.

Otherwise, it will look strange with only the points opacity changing.

Is it OK for you?

ahocevar:
I do not see a way to change the opacity of other geometry types, other than recreating all primitives.

schmidtk:
You have to use Primitive#getGeometryInstanceAttributes and modify the color with the new opacity, but keep in mind that function will fail until the primitive is ready to render. We're also not setting the `id` field on GeometryInstance objects we create in `core.js` [here](https://github.com/openlayers/ol3-cesium/blob/master/src/core.js#L478) and [here](https://github.com/openlayers/ol3-cesium/blob/master/src/core.js#L693), which we should be doing. That will also cause errors in the uncompiled version.

ahocevar:
This is what I tried, but after rendering the geometry instances of all primitives are undefined, so nothing can be changed.

gberaudo:
@ahocevar, here is a working Sandcastle example
```js
Cesium.Math.setRandomNumberSeed(1234);
var viewer = new Cesium.Viewer('cesiumContainer');
var scene = viewer.scene;
var primitives = scene.primitives;
var solidWhite = Cesium.ColorGeometryInstanceAttribute.fromColor(Cesium.Color.WHITE);

var id = 'anid';
var rectangle = Cesium.Rectangle.fromDegrees(-92.0, 20.0, -86.0, 27.0);
var rectangleOutlineInstance = new Cesium.GeometryInstance({
    geometry : new Cesium.RectangleOutlineGeometry({
        rectangle : rectangle
    }),
    attributes : {
        color : solidWhite
    },
    id: id
});


var ro = primitives.add(new Cesium.Primitive({
    geometryInstances : [rectangleOutlineInstance],
    appearance : new Cesium.PerInstanceColorAppearance({
        flat : true,
        translucent : true,
        renderState : {
            lineWidth : Math.min(4.0, scene.maximumAliasedLineWidth)
        }
    })
}));

ro.readyPromise.then(function() {
    var i = 0;
    setInterval(function() {
        var attrs = ro.getGeometryInstanceAttributes(id);
        //var color = attrs.color;
        attrs.color[3] = i;
        attrs.color = attrs.color;
        i = i + 20;
        if (i > 255) {
            i = 0;
        }
    }, 100);
});
```

ahocevar:
I had found a similar example and implemented accordingly. As I said, the problem is that the primitive seems to forget about its geometry instances. I can push my non working code if you want to take a look.

gberaudo:
@ahocevar, Cesium deletes the reference to the geometry instances for
performance reasons.

As suggested by @schmidtk, we need to set and keep a list of the
geometry instance ids.

Please push your code somewhere and I will have a look.

ahocevar:
I just saw that we're not assigning an array to geometryInstances, but a single geometry instance. This is the cause of the problem.

ahocevar:
> I just saw that we're not assigning an array to geometryInstances, but a single geometry instance. This is the cause of the problem.

At least it is part of the problem. The correct approach is probably really to assign ids.

ahocevar:
@gberaudo This is my code so far: https://github.com/openlayers/ol3-cesium/commit/3206bf7f174793d183c47946a4e3b5a141802760

schmidtk:
Do you know if Cesium uses the geometry instance id for anything other than lookup with `getGeometryInstanceAttributes`? If not, we can reuse a constant id instead of maintaining a list.

gberaudo:
@ahocevar, have a look to https://github.com/gberaudo/ol3-cesium/commit/84f566dc45ddb5c15951faead8ba8ea9f8978e29 for how to modify opacity after the primitive is rendered.

ahocevar:
Thanks @gberaudo, that works well. I'll try to complete this pull request when I get to it. Might be a long time from now though.

PR #169, opened 2015-02-12
Use Ol 3.2 and Cesium 1.6

gberaudo:
Migrate to new stable versions of Ol3 and Cesium.


ahocevar:
Looks good, please merge.

gberaudo:

PR #173, opened 2015-02-23
Use pitch and not removed apis

gberaudo:
Tilt is deprecated


elemoine:
+1 from me.

gberaudo:
Thanks for the review @elemoine.

PR #175, opened 2015-02-25
Generate example-list.js earlier

gberaudo:
The example-list.js file was generated after the copy of the
example repository to dist directory.


gberaudo:
Merging trivial fix.

PR #180, opened 2015-03-18
Core function for computing bounding box

gberaudo:
This PR adds a core function for computing a bounding box at a target point.
It takes into account the terrain and may be used for creating a WFS query...

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -24,6 +24,54 @@ goog.require('olcs.core.OlLayerPrimitive');
    */
   olcs.core.glAliasedLineWidthRange = -1;
 
+
+  /**
+   * Compute the pixel width and height of a point in meters using the
+   * camera frustum.
+   * @param {!Cesium.Scene} scene
+   * @param {!Cesium.Cartesian3} target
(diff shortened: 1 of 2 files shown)


src/core.js
```diff
@@ -24,6 +24,56 @@ goog.require('olcs.core.OlLayerPrimitive');
...
+  /**
+   * Compute bounding box around a target point.
+   * @param {!Cesium.Scene} scene
+   * @param {!Cesium.Cartesian3} target
+   * @param {number} amount Half the side of the box, in pixels.
+   * @return {Array<Cesium.Cartographic>} bottom left and top right
+   * coordinates of the box
+   */
+  olcs.core.computeBoundingBoxAtTarget = function(scene, target, amount) {
+    var pixelSize = olcs.core.computePixelSizeAtCoordinate(scene, target);
```
fredj: `pixelSize`may be undefined. Add a check ?

fredj:
LGTM

src/core.js
gberaudo: Fixed, `pixelSize` may never be undefined.

gberaudo:

PR #182, opened 2015-03-23
Fix ol3cs.setEnabled

gberaudo:
Disabling a disabled ol3cs is an error and is now avoided.
Make the API more regular by requiring a parameter.


fredj:
+1

gberaudo:

ahocevar:
@gberaudo, for API breaking changes like this, please also include a note in `CHANGES.md` in the pull request.

Since you merged this pull request already without such a note, please create a follow-up pull request.

gberaudo:
@ahocevar, you are totally right. Please see #183.
Once it is merged, I am going to prepare the March release.

PR #184, opened 2015-03-23
Update CHANGES.md with release date

gberaudo:


CHANGES.md
```diff
@@ -1,5 +1,7 @@
 # Changelog
 
+## v1.3 - 2015-03-23
+
 * Port to Cesium 1.7.1 and Ol 3.3.0
 
 # Breaking changes
```
ahocevar: You'll want a different heading type for the breaking changes (e.g. `###`, or just a bullet point like for previous releases).

ahocevar:
One comment as line note above. Can be merged when fixed.

gberaudo:
You have a sharp eye @ahocevar, thank you for spotting this. I also moved up the breaking change section. I think we should always have it first, is it OK for you?

ahocevar:
Fine with me. Looks good now.

Issue #188, opened 2015-04-17
olcs.RasterSynchronizer is not exported


gejgalis:
I wanted to setup custom createSynchronizers function with only RasterSynchronizer but it is not exported like VectorSynchronizer.

gberaudo:
Thank you @gejgalis for the report, it is now fixed.

PR #189, opened 2015-04-17
Mark RasterSynchronizer @api

gberaudo:
See #188.


gberaudo:
Merging it directly since it is trivial.

PR #194, opened 2015-05-07
Depend on OL 3.5 and Cesium 1.9

gberaudo:


gberaudo:
Thanks for any review.

.gitignore
```diff
@@ -2,3 +2,4 @@
 dist
 node_modules
 examples/example-list.js
+*.swp
```
ahocevar: Shouldn't this be in your local gitignore file?

ahocevar: Or are the `*.swp` files really generated by the build process?

gberaudo: OK, I removed this line and instead created a global gitignore file on my machine like explained in https://help.github.com/articles/ignoring-files/.

ahocevar:
Thanks! Please merge.

gberaudo:
Thank you for the review @ahocevar.

PR #201, opened 2015-06-03
Fix tilegrid issue

gberaudo:
Require https://github.com/openlayers/ol3/pull/3759

### DIFF
--- a/src/core/olimageryprovider.js
+++ b/src/core/olimageryprovider.js
@@ -2,7 +2,7 @@ goog.provide('olcs.core.OLImageryProvider');
 
 goog.require('goog.events');
 goog.require('ol.proj');
-goog.require('ol.tilegrid.XYZ');
+goog.require('ol.tilegrid');
 
 
 
@@ -42,6 +42,13 @@ olcs.core.OLImageryProvider = function(source, opt_fallbackProj) {
   this.emptyCanvas_ = goog.dom.createElement(goog.dom.TagName.CANVAS);
   this.emptyCanvas_.width = 1;
   this.emptyCanvas_.height = 1;
+
+  /**
+   * @type {function(ol.TileCoord, ol.proj.Projection, ol.TileCoord=):
+   * ol.TileCoord}
+   * @private
+   */
+  this.transform_ = goog.functions.identity;
 
   this.source_.on(goog.events.EventType.CHANGE, function(e) {
     this.handleSourceChanged_();
@@ -148,6 +155,16 @@ olcs.core.OLImageryProvider.prototype.handleSourceChanged_ = function() {
         olcs.core.OLImageryProvider.createCreditForSource(this.source_);
     this.credit_ = !goog.isNull(credit) ? credit : undefined;
 
+    var tilegrid = this.source_.getTileGrid();
+    if (goog.isNull(tilegrid)) {
+      this.transform_ = goog.functions.identity;
+    } else {
+      this.transform_ = /** @type {
+          function(ol.TileCoord, ol.proj.Projection, ol.TileCoord=):
+          ol.TileCoord} */ (
+          tilegrid.createTileCoordTransform());
+    }
+
     this.ready_ = true;
   }
 };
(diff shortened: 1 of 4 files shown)


gberaudo:
@ahocevar, thank you for pointing me to `ol.tilegrid.createTileCoordTransform()`.
I tested the rastersync.html example and it behaves correctly except with WMS tile (click remove all layers and then add TileWMS).
It is a regression; do you have an idea of what could be wrong?

ahocevar:
Hm. For WMS the tile coord should not matter at all. Also, the tile coord transform should return the tile coord unaltered. Is that correct?

I'll be offline for the rest of the week, but I'll look at it next week if you cannot figure it out.

src/core/olimageryprovider.js
```diff
@@ -208,12 +208,9 @@ olcs.core.OLImageryProvider.prototype.requestImage = function(x, y, level) {
...
   if (!goog.isNull(tileUrlFunction) && !goog.isNull(this.projection_)) {
     // perform mapping of Cesium tile coordinates to ol3 tile coordinates
-    var z_ = (this.tilingScheme_ instanceof Cesium.GeographicTilingScheme) ?
-             (level + 1) : level;
-    var y_ = (this.source_.getTileGrid() instanceof ol.tilegrid.XYZ) ?
-             y : (y - (1 << level));
-    y_ = -y_ - 1; // opposite indexing
-    var url = tileUrlFunction([z_, x, y_], 1, this.projection_);
+    var transform = this.source_.getTileGrid().createTileCoordTransform();
+    var tileCoord = transform([level, x, y]);
```
ahocevar: You have to do the transform on the original tileCoord, before changing the level. This might also explain your issue with TileWMS.

src/core/olimageryprovider.js
```diff
@@ -208,12 +208,9 @@ olcs.core.OLImageryProvider.prototype.requestImage = function(x, y, level) {
   var tileUrlFunction = this.source_.getTileUrlFunction();
   if (!goog.isNull(tileUrlFunction) && !goog.isNull(this.projection_)) {
     // perform mapping of Cesium tile coordinates to ol3 tile coordinates
-    var z_ = (this.tilingScheme_ instanceof Cesium.GeographicTilingScheme) ?
-             (level + 1) : level;
-    var y_ = (this.source_.getTileGrid() instanceof ol.tilegrid.XYZ) ?
-             y : (y - (1 << level));
-    y_ = -y_ - 1; // opposite indexing
-    var url = tileUrlFunction([z_, x, y_], 1, this.projection_);
+    var transform = this.source_.getTileGrid().createTileCoordTransform();
```
ahocevar: Better to cache the transform, instead of creating it with every call.

gberaudo:
Indeed, the issue should come from elsewhere. I will have a look.

src/core/olimageryprovider.js
gberaudo: I implemented your suggestion but had to cast the result from `this.source_.getTileGrid().createTileCoordTransform();`. I guess calling a member function is somehow incorrect/unsupported and we should access it through the prototype. I will add it to Ol3.

gberaudo:
The WMS regression seems to have been introduced by Ol3 commit https://github.com/openlayers/ol3/commit/b05193fa454a963b33e6c5965181abde9b45b7fa.
I created an ol3-cesium branch to help test: https://github.com/gberaudo/ol3-cesium/tree/wip_testing_wms_regression. @ahocevar, would you mind having a look to this?
Before this commit, the WMS layer appears on the Cesium globe; after this commit, the globe stays blue.

ahocevar:
I'll take a look @gberaudo.

ahocevar:
@gberaudo I added two commits that fix the issue: https://github.com/gberaudo/ol3-cesium/compare/wip_port_ol3_master...ahocevar:wip_port_ol3_master. Note that v3.6.0 of ol3 is needed.

I'm going to create another branch that depends on https://github.com/ahocevar/ol3/tree/call-tileurlfunction-with-transformed-tilecoord. With that, we can just call the tileUrlFunction without transforming/modifying anything.

gberaudo:
Hi @ahocevar, thank you for that.

I updated the PR with your work but there is still an issue when the
tilegrid is null. In this case, it seems the transform should not be
assumed to be the identity.

I am looking forward to your other branch.

ahocevar:
@gberaudo I put the commit that fixes everything in the same branch: 5d30725128a1c9b295883b3f23cbe5ceb8b3d765.

There is still a mapping necessary. I had forgotten that ol3 uses bottom-left origin tile coordinates for calling the ol.source.TileWMS tileUrlFunction by default.

gberaudo:
@ahocevar, thank you for stepping in and fixing this regression.
Once your PR is mergeid in OL3 I will update the OL3 submodule and merge this PR.

gberaudo:
Now that your OL3 PR is merged I can update the dependency to OL3 master.
However, I will wait a bit to avoid beeing hit by the removal of featureOverlay.

ahocevar:
I just proposed another pull request, openlayers/ol3#3815, which allows us to remove the conditional tile coord transformation. I updated my branch, see 0b25b5450da5eef994a79bfe36858f7254626d69.

ahocevar:
@gberaudo:

> However, I will wait a bit to avoid beeing hit by the removal of featureOverlay.

Should not be an issue, because ol3-cesium does not rely on FeatureOverlay anywhere, and also has no support for it.

gberaudo:
Thank you @ahocevar for having worked on this. I cherry picked your commit 0b25b54 and will now merge.

ahocevar:
Thanks for bringing this in @gberaudo.

PR #202, opened 2015-06-18
Port to cesium 1.10 and minor improvements

gberaudo:
And also:
- change serving port from 3000 to 4000 to allow running both ol3 and ol3-cesium concurrently;
- add some documentation;
- add or update some Cesium externs.


fredj:
LGTM

gberaudo:
Thanks for the review @fredj.

PR #203, opened 2015-06-18
Point on terrain

gberaudo:


gberaudo:
Any review?

src/core.js
```diff
@@ -820,6 +823,7 @@ goog.require('olcs.core.OlLayerPrimitive');
    */
   olcs.core.olPointGeometryToCesium = function(geometry, projection, style,
       billboards, opt_newBillboardCallback) {
+    var altitudeMode = geometry.get('altitudeMode');
```
ahocevar: Not sure if it's good to rely on a per-geometry object setting for this. The workflow for clamping all geometries of all features to the ground is a bit weird then from an API perspective. I think it would be better to make this an OLCesium instance option.

gberaudo: Since the code sprint at Schladming, Ol3 is able to parse the `altitudeMode` from KML geometries and set them as a geometry properties of Ol3 features.

In this PR I simply read this property and honour the requested mode.

I agree it would be great in the future to also handle `altitudeMode` at the layer level but it is not possible with the current design of the converters. I think we can add it later.

ahocevar: Oh, I was not aware that the KML change we discussed got merged. In this case it makes sense I think. So please merge after adding some documentation about the new feature to olcs.VectorSynchronizer. You know: if a feature is not documented, it's not there :smile: .

gberaudo:
Thank you for the review @ahocevar. I added documentation in the changelog about the new `scene` parameter for `olcs.core.OlLayerPrimitive` and `olcs.core.olVectorLayerToCesium()`.

PR #204, opened 2015-06-19
Feature converters as a class

gberaudo:
With this change, a user can easily override any step of the
conversion process by subclassing the base class.

Subclassing requires that the subclass and the library be compiled together using
the closure compiler. This due to how `@export` work.

It allows application developers to implement custom behaviours like https://github.com/openlayers/ol3-cesium/issues/164#issuecomment-113253301.


gberaudo:
I added some related commits:
Now passing more context to the conversion methods to allow more powerful subclassing.
Enhance picking experience.
Also consider `altitudeMode` from the feature and layer.

See https://github.com/openlayers/ol3-cesium/issues/164#issuecomment-113253301 and https://github.com/openlayers/ol3-cesium/pull/203#discussion_r32736536

gberaudo:
Following an internal discussion with @elemoine and @fredj, it appears that subclassing requires compiling the subclass and the library together. As a consequence, I updated the PR so that this requirement is clear to the users.
@ahocevar, @elemoine, @fredj, do you have other comments?

ahocevar:
Documentation in the upgrade notes will not be visible to a typical user, unless they upgrade from a previous version. So this conversion process should be documented in a way that it appears in the API docs.

I am also concerned about the requirement to compile application/subclasses and library together. This could be avoided by making those functions that can be overriden part of the constructor options in `externs/olcsx.js? . Then you could do something like
```js
new ol.FeatureConverter({
  scene: scene,
  olLineStringGeometryToCesium: function(layer, feature, olGeometry, projection, olStyle) {
    // ...
  }
});

**Comment by gberaudo** [comment]
@ahocevar, I added documentation to the `olcs.FeatureConverter` constructor:
```
/**
 * Concrete base class for converting from OpenLayers3 vectors to Cesium
 * primitives.
 * Extending this class is possible provided that the extending class and
 * the library are compiled together by the closure compiler.
 * @param {!Cesium.Scene} scene Cesium scene.
 * @constructor
 * @api
 */
olcs.FeatureConverter = function(scene);
```
What do you suggest I should add?

Regarding the requirement to compile application and subclass together with the library, I know about the hack of passing functions to the constructor but it is not the standard way of doing inheritance and it will feel unnatural to all application developers.

An easy and much cleaner solution is that the application developers not using the closure compiler fork the library and implement the subclasses they need, compile and integrate the resulting custom library in their application. This way, we can keep the code simple and natural for everyone.

**Comment by ahocevar** [comment]
> I know about the hack of passing functions to the constructor

This is not a hack. We do it in OpenLayers too, e.g. the `render` option in `ol.Control`.

**Comment by gberaudo** [comment]
While passing a few callback closures to a constructor is perfectly fine, there are 13 overridable `@api` methods and 7 overridable `@protected` methods in the `olcs.FeatureConverter` class. I don't imagine using the closure technique in this case and that is why I consider it would be a `hack` in this particular case.

**Comment by ahocevar** [comment]
You meant to allow overriding `@protected` methods too, not just `@api`? In this case the first thing to do would be to think about a sensible and simple API and reduce the number of hooks - no average user will be able to make use of such complexity.

At least I understand now why you're saying this only works when compiling the overrides together with the library. Because with everything that is `@api`, this would not be required.

**Comment by gberaudo** [comment]
This PR simply moves the conversion static functions to a class in order to allow customization
of the conversion logics. It allows all previous use cases and enable new possibilities when you compile with closure.

What were public conversion functions are now public methods: there are 13 of these.
Anonymous low level functions are now marked protected: there are 7 of these. I can remove the `@protected` annotation if you feel it is better, it is very minor to me.

We need at least one public method per geometry type, plus some layer or low-level methods: I do not see how we could reduce these to a few closure hooks.

Default behaviour should be what the users need. In rare cases, a user may override one or two methods. Users needing to change all methods should consider contributing a patch.
I think it is perfectly manageable.

Regarding the requirement to compile with closure, it also holds for methods with the `@api` annotations. Recall that closure work with minified names in `advanced` mode and that the export is only an entrance point.
This issue would not occur in `simple` and nothing prevents a user to compile the library in `simple` mode and then directly extend any class in application code.

**Comment by ahocevar** [comment]
In this case I do not see the advantage that users gain from this pull request. By the same means (compile overrides with the library), users can already override the static conversion functions that we have now.

Or do you have a specific use case where you do not want to override any of these static functions globally?

**Comment by gberaudo** [comment]
What do you mean @ahocevar?
With current core static functions, how would you override the `olcs.core.olLineStringGeometryToCesium` function and how would you make it used by the `olcs.core.olMultiGeometryToCesium` function?

**Comment by ahocevar** [comment]
Mabe your use case is differnt, but you could always create an interceptor:
```js
olcs.core.olLineStringGeometryToCesium = (function() {
  var orig = olcs.core.olLineStringGeometryToCesium;
  return function(olGeometry, projection, olStyle) {
    // do something
    var primitives = orig(olGeometry, projection, olStyle);
    // do something with primitives
    return primitives;
  };
})();
```

The problem I see with removing these static functions is that you can currently use them wherever you want, also outside of a FeatureConverter or whatever. When they are members of an instance, you cannot easily use them outside of an instance any more.

Also note that these static functions were an initial design principle and requirement of Camptocamp, so maybe @elemoine also has an opinion on this change.

gberaudo:
@ahocevar, I think using an interceptor would not work:
Imagine the `olcs.core.olLineStringGeometryToCesium` was renamed to `olcs.core.S` by closure and exported with a code like `olcs.core.olLineStringGeometryToCesium = olcs.core.S`.
Then the interceptor snippet will not be enough since the `olcs.core.S` will still hold the original function and the `olcs.core.olMultiGeometryToCesium` will not call in our wrapper.

Regarding the use of core functions, if you recall our initial design discussions, the goal was to allow flexibility and reuse. As we can all see now, we can do better in this case.

A static function is basically a global function. It is always possible to migrate existing code by introducing a singleton and calling through it: `app.converter = new olcs.FeatureConverter({scene: scene});`. Then simply change `olcs.core.olLineStringGeometryToCesium()` to `app.converter.olLineStringGeometryToCesium()`.

@ahocevar, does it answer your concerns?

ahocevar:
I don't have strong concerns about the pull request. Just wanted to see if this change is in line with Camptocamp's initial requirements. If it is, I'm fine with it. But please do add the snippets from your last paragraph (the ones about the singleton use) above to the upgrade notes, because this change affects some ol.core `@api` functions. When done, and when @elemoine is also fine with the change, you can merge.

gberaudo:
@ahocevar, I updated the `CHANGES.md` as you requested and moved out two low level functions.

src/core.js
```diff
@@ -497,15 +490,16 @@ goog.require('olcs.core.OlLayerPrimitive');
...
+   * Reproject an OpenLayers geometry to EPSG:4326 if needed.
+   * The geometry will be cloned only when original projection is not EPSG:4326
+   * and the properties will be shallow copied.
    * @param {!T} geometry
    * @param {!ol.proj.ProjectionLike} projection
    * @return {!T}
    * @template T
+   * @api
    */
-  var olGeometryCloneTo4326 = function(geometry, projection) {
```
elemoine: With this removed the wrapping in a self-called function may not be necessary anymore.

gberaudo: You have sharp eyes @elemoine. It is indeed a good opportunity to remove it and it will
be valuable when testing. I just commited this internal change.

gberaudo:
I fixed the converter which was not honouring polyline widths, removed the self-calling function in core.js, added a missing extern and an overridable method for manipulating billboard option.
@elemoine, is it OK for you if I merge this PR?

elemoine:
As I said to @gberaudo already I just find it regrettable that extending the `FeatureConverter` class is limited to users using Closure Compiler. That being said, I am not opposed to merging this PR, as the issue can be addressed in the future, when the problem comes up.

gberaudo:
Thank you @elemoine for taking time for reviewing this PR.
I agree with you regarding the requirement of the closure compiler for extending this new `olcs.FeatureConverter` class and also the `olcs.AbstractSynchronizer`, `olcs.VectorSynchronizer` and `olcs.RasterSynchronizer` classes.
I will merge the PR now and in the future we can add a setup for these users.

PR #208, opened 2015-06-30
Set Ol3 references for picking to all primitives

gberaudo:


fredj:
LGTM

gberaudo:

Issue #212, opened 2015-07-30
olcs.Camera readFromView() as public API


gejgalis:
Please mark this method to be available in public API

gberaudo:
It is conceivable to mark the method API but what is your use case?

gejgalis:
I'm using  standalone `var camera = new olcs.Camera(scene, map)` and I'm synchronizing manually both views.

gberaudo:
I understand, could you please create a pull request adding the `@api` annotation?

gberaudo:
Implemented in #216.

PR #216, opened 2015-08-04
Mark olcs.Camera#readFromView @api

gejgalis:
Fix for issue #212


gberaudo:
Thank you for the patch @gejgalis.

Issue #213, opened 2015-07-31
Cesium.ImageryProvider#pickFeatures method is not implemented


gejgalis:
OpenLayers: `3.4`
ol3-cesium: `1.6`
Cesium: `1.11`

When working with Cesium.Viewer and ol3-cesium there is an error thrown by Cesium when user clicks twice on earth (picking features from globe) about not implemented method `pickFeatures` in `Cesium.ImageryProvider`.

Please add following getter into `olcs.core.OLImageryProvider`:

`pickFeatures: {get: function () {return undefined}}`

gberaudo:
I tested with http://openlayers.org/ol3-cesium/examples/synthvectors.html which is using Cesium 1.10 and could not reproduce the problem you describe.

What do you mean with `When working with Cesium.Viewer`?

gejgalis:
I'm just using Camera and RasterSynchronizer with Cesium.Viewer. VectorSynchronizer doesn't meet our different requirements for 2D and 3D mode. Moreover I'm using Cesium Entity API to populate 3D features on map, pick them and other easy to use functionalities from Cesium.Viewer.

Double click on globe tries to pick Imagery, that's why I could see this error.

```javascript
...
var viewer = new Cesium.Viewer($targetEl, {
    scene3DOnly: true,
    animation: false,
    geocoder: false,
    baseLayerPicker: false,
    infoBox: true,
    fullscreenButton: false,
    homeButton: false,
    sceneModePicker: false,
    selectionIndicator: true,
    timeline: false,
    navigationHelpButton: false,
    navigationInstructionsInitiallyVisible: false,
    imageryProvider: false
});

this.syncCamera = new olcs.Camera(viewer.scene, map);
this.syncRaster = new olcs.RasterSynchronizer(map, viewer.scene);
...
```

gberaudo:
OK, I understand now. Could you create a pull request adding the `pickFeatures` method like in
https://github.com/AnalyticalGraphicsInc/cesium/blob/1.11/Source/Scene/BingMapsImageryProvider.js#L534 ?

Regarding the vector synchronizer, do you have examples about what you think is missing?

gejgalis:
Our map basically shows airports and aircrafts with their routes. In 2D map airports are painted as circles with triangles inside (style function return 2 styles). In 3D mode we shows airports as pins.
Routes in 2D are lines (sometimes dotted) and walls in 3D.

[...]

PR #217, opened 2015-08-04
Added optional function pickFeatures 

gejgalis:
Fix for #213


gberaudo:
Thank you for the patch @gejgalis.

PR #219, opened 2015-08-06
Fix removal of billboard in the absence of a geometry

gberaudo:
When doing feature.setGeometry(null) on a Point feature, the Cesium
counterpart (Billboard) was not removed properly.

This PR handles this case.
The case when there is no geometry and no associated billboard is found
for this feature id is also handled.

### DIFF
--- a/src/vectorsynchronizer.js
+++ b/src/vectorsynchronizer.js
@@ -96,14 +96,12 @@ olcs.VectorSynchronizer.prototype.createSingleCounterpart = function(olLayer) {
   var onRemoveFeature = goog.bind(function(feature) {
     var geometry = feature.getGeometry();
     var id = goog.getUid(feature);
-    if (goog.isDefAndNotNull(geometry) && geometry.getType() == 'Point') {
+    if (!geometry || geometry.getType() == 'Point') {
       var context = csPrimitives.context;
-      var bbs = context.billboards;
       var bb = context.featureToCesiumMap[id];
       delete context.featureToCesiumMap[id];
-      if (goog.isDefAndNotNull(bb)) {
-        goog.asserts.assertInstanceof(bb, Cesium.Billboard);
-        bbs.remove(bb);
+      if (bb instanceof Cesium.Billboard) {
+        context.billboards.remove(bb);
       }
     }
     var csPrimitive = featurePrimitiveMap[id];


src/vectorsynchronizer.js
```diff
@@ -96,7 +96,7 @@ olcs.VectorSynchronizer.prototype.createSingleCounterpart = function(olLayer) {
   var onRemoveFeature = goog.bind(function(feature) {
     var geometry = feature.getGeometry();
     var id = goog.getUid(feature);
-    if (goog.isDefAndNotNull(geometry) && geometry.getType() == 'Point') {
+    if (!geometry || geometry.getType() == 'Point') {
```
elemoine: What if the feature being removed previously had a geometry that was not a point? Isn't the `goog.asserts.assertInstanceof(bb, Cesium.Billboard)` assertion going to fail in that case?

gberaudo: You are right @elemoine, this is now fixed.

src/vectorsynchronizer.js
```diff
@@ -96,14 +96,15 @@ olcs.VectorSynchronizer.prototype.createSingleCounterpart = function(olLayer) {
...
-      var bbs = context.billboards;
       var bb = context.featureToCesiumMap[id];
       delete context.featureToCesiumMap[id];
       if (goog.isDefAndNotNull(bb)) {
-        goog.asserts.assertInstanceof(bb, Cesium.Billboard);
-        bbs.remove(bb);
+        var bbs = context.billboards;
+        if (bb instanceof Cesium.Billboard) {
+          bbs.remove(bb);
+        }
```
elemoine: I think you can simplify the code:

Instead of

```js
  if (goog.isDefAndNotNull(bb)) {
    var bbs = context.billboards;
    if (bb instanceof Cesium.Billboard) {
      bbs.remove(bb);
    }
  }
```

you can use:

```js
  if (bb instanceof Cesium.Billboard) {
    context.billboards.remove(bb);
  }
```

src/vectorsynchronizer.js
gberaudo: And the code is now more clear, thanks @elemoine.

elemoine:
LGTM.

gberaudo:

PR #221, opened 2015-08-07
Relax geometry type of csAddBillboard

gberaudo:
Allow other geometry types to be converted to billboards.


elemoine:
+1

gberaudo:
Thank you for the review @elemoine. Merging this trivial patch.

PR #228, opened 2015-08-14
Allow experimental lazy loading of Cesium.js

gberaudo:
Experimental feature to allow loading the 3D part only when it is
needed. See the 'lazy' example.

This commit uncouple OL3-Cesium inheritance chains depending on Cesium
until the objects are actually instanciated.


src/core/ollayerprimitive.js
```diff
@@ -10,7 +10,11 @@ goog.provide('olcs.core.OlLayerPrimitive');
  * @extends {Cesium.PrimitiveCollection}
  */
 olcs.core.OlLayerPrimitive = function(layerProjection, scene) {
-  goog.base(this);
+  if (!olcs.core.OlLayerPrimitive.initialized) {
+    olcs.core.OlLayerPrimitive.initialized = true;
+    goog.inherits(olcs.core.OlLayerPrimitive, Cesium.PrimitiveCollection);
+  }
+  Cesium.PrimitiveCollection.call(this);
```
elemoine: Out of curiosity, any reason for not using `goog.base` anymore?

elemoine: I see the related comment in the other file, but I still don't understand.

gberaudo: In this case `goog.base()` will just be replaced by the call to the super constructor.
This change is required due to the move of the call to `goog.inherits()` in order to workaround a verification of the closure compiler.

src/core/ollayerprimitive.js
```diff
@@ -10,7 +10,15 @@ goog.provide('olcs.core.OlLayerPrimitive');
...
  */
 olcs.core.OlLayerPrimitive = function(layerProjection, scene) {
-  goog.base(this);
+  if (!olcs.core.OlLayerPrimitive.initialized_) {
+    olcs.core.OlLayerPrimitive.initialized_ = true;
+    goog.inherits(olcs.core.OlLayerPrimitive, Cesium.PrimitiveCollection);
+  }
+
+  // Directly call the super constructor instead of using
+  // goog.base to avoid a check for goog.inherit() by the compiler.
```
elemoine: `inherit` -> `inherits`

elemoine:
Not a huge fan of this, as it sounds hackish to me. But I understand the need and I don't have anything to propose. So I agree with merging this if @ahocevar agrees as well.

src/core/ollayerprimitive.js
gberaudo: I added comments to the code about the direct call to the super constructor and the '`Cesium.ImageryProvider` interface'.

ahocevar:
This looks hackish to me as well. Instead of lazily loading Cesium, couldn't you just load olcs lazily instead?

ahocevar:
Another option for the ImageryProvider would be to not use inheritance at all, not even a fake one. The interface in Cesium is only there to describe what a custom ImageryProvider needs to provide. See https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Scene/TileCoordinatesImageryProvider.js for an example of a custom ImageryProvider. So you can create an OlImageryProvider without even referencing the Cesium.ImageryProvider interface.

In the case of the LayerPrimitive, you could create a factory function that creates a plain Cesium.PrimitiveCollection instance, mixes in everything that olcs needs, and returns the instance.

gberaudo:
Javascript is a `prototype` language; I would not call `hackish the fact of manipulating the prototype chain like `goog.inherits()` does.

On the contrary, `goog.base()` is black magic and is using `callee.caller` which has been removed in ES5 strict.

Delaying the call to `goog.inherits()` to only manipulate the prototype chain when it is needed seems reasonable to me. I am missing something?

Regarding `ImageryProvider`, I previously removed all manipulation of the inheritance chain so there is no more issue with it.

ahocevar:
By hackish I mean that you build the inheritance chain inside the constructor of the subclass.

I may be missing your use case, but why don't you load olcs lazily instead?

elemoine:
> I may be missing your use case, but why don't you load olcs lazily instead?

Because it may be compiled together with the application code, which also includes 2D code that is be loaded right away at page init time.

elemoine:
@gberaudo it's hackish because you use `goog.inherits` in a non-standard way. A proof being that the compiler complains when you use `goog.inherits` this way and call `goog.base`. It's not standard so you'll have to explain to olcs developers why they need to do things this way when extending Cesium classesro

elemoine:
I am personally not opposed to that change, just trying to understand if there's a better way to address the "deferring Cesium loading" use-case.

ahocevar:
Wouldn't this be an acceptable solution then, @elemoine & @gberaudo?

> In the case of the LayerPrimitive, you could create a factory function that creates a plain Cesium.PrimitiveCollection instance, mixes in everything that olcs needs, and returns the instance.

gberaudo:
@elemoine, `goog.base()` has known limitations and in this case replacing it with a standard and more robust solution is straightforward. Once done, nothing prevents using `goog.inherits()` lazily.

I do not think forbidding OL3-Cesium to extend Cesium types is the right way to go: it would be a strong limitation for the project which we will certainly regret in the future.

A long term solution to the lazyloading problem seems to be the `--module` parameter of the closure compiler, which allows to output several lazyloadable modules. However, the implementation would be complex and probably require heavy changes to closure-util. Moreover, it would certainly be more involved for the users.

This complexity is the reason why I favored the proposed solution, which is straightforward.

elemoine:
> I do not think forbidding OL3-Cesium to extend Cesium types is the right way to go: it would be a strong limitation for the project which we will certainly regret in the future.

Do you anticipate a lot of cases where we'd need to inherit from Cesium classes? Is inheriting from Cesium classes a common pattern when using Cesium?

gberaudo:
Good question.
Cesium is more about duck typing than inheritance. However, there may still be other cases where we will want to extend / wrap a Cesium type in our own type.
With a pure duck typing approach we would have to duplicate Cesium code and logics which are likely to change.

elemoine:
+1 for merging this if you think forbidding inheriting from Cesium classes is not an option for ol3-cesium.

ahocevar:
Have you considered something like this instead:
```js
olcs.core.OlLayerPrimitive = function(layerProjection, scene) {
  if (goog.isDef(Cesium) && goog.isFunction(Cesium.PrimitiveCollection) {
    olcs.core.OlLayerPrimitive = function(layerProjection, scene) {
      goog.base(this);
      // ...
    }
    goog.inherits(olcs.core.OlLayerPrimitive, Cesium.PrimitiveCollection);

    return new olcs.core.OlLayerPrimitive(layerProjection, scene);
  }
  goog.asserts.assert(goog.isDef(Cesium), 'Cesium is not loaded');
};
```
Here you create the OlLayerPrimitive class with inheritance as it is meant, but when calling it for the first time, it will create the prototype before creating the instance.

gberaudo:
@elemoine, @ahocevar, I reworked `olcs.core.OlLayerPrimitive`.
The new type is less easy to use but it is manageable and there is no more inheritance.
The previously proposed technique may still be necessary in the future, but with this commit we can skip it for now. Do you prefer this solution?

src/core/vectorlayercounterpart.js
```diff
@@ -1,4 +1,4 @@
-goog.provide('olcs.core.OlLayerPrimitive');
+goog.provide('olcs.core.VectorLayerCounterpart');
```
elemoine: I don't like this name. I think I'd just name it `olcs.core.LayerPrimitive`.

src/core/vectorlayercounterpart.js
```diff
@@ -25,6 +23,14 @@ olcs.core.OlLayerPrimitive = function(layerProjection, scene) {
...
-  this.add(billboards);
+  this.rootCollection_.add(billboards);
+  this.rootCollection_.add(primitives);
+};
+
+
+/**
+ * @return {!Cesium.Primitive}
+ */
+olcs.core.VectorLayerCounterpart.prototype.getRootPrimitive = function() {
```
elemoine: I'd use a more explicit name: `getCesiumPrimitiveCollection`.

elemoine:
> Do you prefer this solution?

YES. I just don't like the naming, as indicated in my comments to the commits.

src/core/vectorlayercounterpart.js
gberaudo: We are not heriting from `Cesium.Primitive` anymore so it would be misleading to keep the `Primitive` suffix. Moreover, this type is specific to Vector layers: skipping the keyword would be ambiguous.

src/core/vectorlayercounterpart.js
gberaudo: I am returning a `Cesium.Primitive` and not a `Cesium.PrimitiveCollection` on purpose and the name was chosen accordingly. I like the `root` which suggests clearly that the hierarchy of Cesium primitives is hidden and should be considered a black box. I could have chosen `getRootCesiumPrimitive` but it sounds redundant.

gberaudo:
@elemoine, I explained the reasons for the namings. Do they make sense for you now?

ahocevar:
This approach is much better for now. I have no strong opinion on the naming. Looks good to me when @elemoine is happy too.

src/core/vectorlayercounterpart.js
elemoine: So I meant `getCesiumPrimitive`.

src/core/vectorlayercounterpart.js
elemoine: I don't think it is misleading to have `Primitive` in the class name personally. It's a matter of taste but I find the name `VectorLayerCounterpart` ugly :) And I find it inexplicit as well. I'd like `olcs.core.PrimitiveWrapper` better than `VectorLayerCounterpart`.

elemoine:
I provided suggestions for the naming. But feel free to merge with or without taking my suggestions into account.

src/core/vectorlayercounterpart.js
gberaudo: `Primitive` is a `Cesium` concept so I mean having both `Cesium` and `Primitive` sounds redundant.
On the contrary there are several primitives stored in the object and that is why it is important to make clear we are returning the `root` of the Primitive hierarchy.
For example, we also store the shared billboard collection in the object and we may add a getter for it in the future.

gberaudo:
Thank you for the review @elemoine and @ahocevar.
It is very great we were able to get this feature into OL3-Cesium.

gberaudo:
Keeping the intermediate commit for the record.

Issue #146, opened 2014-12-17
Spurious use of transpose in abstract synchronizer


gberaudo:
Transpose should only be used when the values are convertible to string keys.
It is not the case in the abstract synchronizer.

```
this.layerMap is
  1: null
  2024: olcs.core.OlLayerPrimitive
  3026: olcs.core.OlLayerPrimitive
  4028: olcs.core.OlLayerPrimitive
  __proto__: Object
```

```
this.unusedCesiumObjects_ is
  [object Object]: "4028"
  null: "1"
  __proto__: Object
```

gberaudo:
Fixed by #241.

PR #241, opened 2015-09-07
Seamless layer synchronization + z-index

gberaudo:
This PR reworks the synchronizers to fix several issues, notably the blue flash when adding/updating and removing layers, and implements Z-index ordering of raster layers.

The abstract synchronizer used to remove all layers from Cesium before any operation (add, update and remove). Now layers are removed only when needed, which prevents the blue flash.

Additionnally, the following issues were fixed:
- buggy transposition of `this.layerMap`;
- missing clean up of vector synchronizer listeners.

### DIFF
--- a/Cesium.externs.js
+++ b/Cesium.externs.js
@@ -1457,6 +1457,12 @@ Cesium.PrimitiveCollection = function() {};
 Cesium.PrimitiveCollection.prototype.add = function(poly) {};
 
 /**
+ * @param {number} index
+ * @return {!Cesium.Primitive}
+ */
+Cesium.PrimitiveCollection.prototype.get = function(index) {};
+
+/**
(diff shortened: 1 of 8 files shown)


gberaudo:
This is a big PR so I propose we keep the commit history (no squash of the commits).

fredj:
Fixes #146

gberaudo:
Any review?

fredj:
`goo.require('goog.array';)` and `goo.require('goog.object';)` should be added in `src/abstractsynchronizer.js` (don't know why the compiler doesn't complain)

fredj:
> `goog.require('goog.array';)` and `goog.require('goog.object';)` should be added in `src/abstractsynchronizer.js` (don't know why the compiler doesn't complain)

https://github.com/openlayers/ol3-cesium/pull/96 and the `ignoreRequires` option

fredj:
+1, very nice improvement !

gberaudo:

PR #243, opened 2015-09-15
GroundPrimitive

oterral:
Use GroundPrimitive instead of Primitive when altittudeMode is set to clampToGround.

Ready for review @gberaudo .

### DIFF
--- a/Cesium.externs.js
+++ b/Cesium.externs.js
@@ -1057,6 +1057,21 @@ Cesium.Primitive.prototype.olFeature;
  */
 Cesium.Primitive.prototype.olLayer;
 
+
+/**
+ * @typedef {{
+ *   geometryInstance: !Cesium.GeometryInstance
+ * }}
+ */
+Cesium.optionsGroundPrimitive;
+
+/**
+ * @constructor
+ * @param {Cesium.optionsGroundPrimitive=} opt_opts
+ * @extends {Cesium.Primitive}
+ */
+Cesium.GroundPrimitive = function(opt_opts) {};
+
 /**
  * @constructor
  */
--- a/src/featureconverter.js
+++ b/src/featureconverter.js
@@ -42,7 +42,8 @@ olcs.FeatureConverter.prototype.setReferenceForPicking =
  * Basics primitive creation using a color attribute.
  * Note that Cesium has 'interior' and outline geometries.
  * @param {ol.layer.Vector} layer
- * @param {!ol.Feature} feature Ol3 feature..
+ * @param {!ol.Feature} feature Ol3 feature.
+ * @param {!ol.geom.Geometry} olGeometry Ol3 geometry.
  * @param {!Cesium.Geometry} geometry
  * @param {!Cesium.Color} color
  * @param {number=} opt_lineWidth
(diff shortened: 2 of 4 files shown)


Cesium.externs.js
```diff
@@ -1421,6 +1436,23 @@ Cesium.PolylineMaterialAppearance = function(object) {};
...
+ * height: (number|undefined),
+ * extrudedHeight: (number|undefined),
+ * vertexFormat: number
+ * }}
+ */
+Cesium.optionsCorridorGeometry;
+
+/**
+ * @constructor
+ * @param {Object=} object
```
gberaudo: Cesium.optionsCorridorGeometry

examples/vectors-clamptoground.html
```diff
@@ -0,0 +1,32 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
gberaudo: I would better not duplicate the vector.js example but enhance it with a button doing:
```js
vectorLayer.set('altitudeMode', 'clampToGround');
vectorLayer2.set('altitudeMode', 'clampToGround');
```

gberaudo:
안녕하세요, @oterral
Nice to see you have started to stab this.

Adding `GroundPrimitive` support and implementing lines on terrain are two different tasks.
The last one needs to be done in Cesium because corridors take meters but we want to use pixels and the conversion depends on the distance. With a fixed ratio, we get the following incorrect reult:
meters_variable_width
pixels_to_meters

See my other comments.

src/featureconverter.js
```diff
@@ -295,21 +309,39 @@ olcs.FeatureConverter.prototype.olLineStringGeometryToCesium =
...
+  var heightReference = this.getHeightReference(layer, feature, olGeometry);
+
+  var outlinePrimitive;
+
+  if (heightReference == Cesium.HeightReference.CLAMP_TO_GROUND) {
+    // Handle both color and width
+    var outlineGeometry = new Cesium.CorridorGeometry({
+      // always update Cesium externs before adding a property
+      positions: positions,
+      width: this.extractLineWidthFromOlStyle(olStyle) * 10000,
```
gberaudo: The conversion factor can not be static, see https://github.com/openlayers/ol3-cesium/pull/243#issuecomment-140387081

examples/vectors-clamptoground.html
oterral: good idea

oterral:
I agree, using Corridor for outline is not ideal. It's workaround until lines on terrain will be implemented.

For the hardcoded factor.  it will be great if we could do like the scaleByDistance  options of [PointPrimitive](https://cesiumjs.org/Cesium/Build/Documentation/PointPrimitive.html?classFilter=Point)

oterral:
@gberaudo I've removed the use of Corridor geometry

gberaudo:
@oterral, could you please modify the vector example as suggested in https://github.com/openlayers/ol3-cesium/pull/243#discussion_r39507298?

oterral:
Yes I'm doing it.

gberaudo:
@oterral, the clampToGround mode button does not seem to work. Maybe it does not work when geometries have absolute 3D coordinates? If it is the case, you could add a new 2D polygon feature to the vector example? See `examples/data/geojson/vector_data.geojson`.

oterral:
Ok I had not tested I was on windows yesterday and nothing worked

oterral:
Ok it works now, there was a typo issue and the layer must be remove then re-add. Is there a way to r-render a layer without adding/removing it?

gberaudo:
Your workaround is OK. I created an issue
https://github.com/openlayers/ol3-cesium/issues/250.

gberaudo:
It looks good now, please merge.

oterral:
Thx for the review but I can't merge I haven't the right :)

gberaudo:
OK, done

PR #246, opened 2015-09-22
Remove goog.array.forEach

gberaudo:
Prefer standard Ecmascript 5.1 array.forEach instead of the function
provided by the closure library. This is part of a work to shrink dependency
on goog library.

### DIFF
--- a/src/core/olimageryprovider.js
+++ b/src/core/olimageryprovider.js
@@ -169,7 +169,7 @@ olcs.core.OLImageryProvider.createCreditForSource = function(source) {
   var text = '';
   var attributions = source.getAttributions();
   if (!goog.isNull(attributions)) {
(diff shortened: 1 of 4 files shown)


src/core/olimageryprovider.js
```diff
@@ -169,8 +169,7 @@ olcs.core.OLImageryProvider.createCreditForSource = function(source) {
   var text = '';
   var attributions = source.getAttributions();
   if (!goog.isNull(attributions)) {
-    goog.array.forEach(attributions, function(el, i, arr) {
-      // strip html tags (not supported in Cesium)
```
fredj: keep the comment ?

fredj:
+1

gberaudo:

PR #247, opened 2015-09-22
Goog cleanup

gberaudo:
Use standard javascript instead.

### DIFF
--- a/src/abstractsynchronizer.js
+++ b/src/abstractsynchronizer.js
@@ -1,6 +1,5 @@
 goog.provide('olcs.AbstractSynchronizer');
 
-goog.require('goog.array');
 goog.require('goog.object');
 
 goog.require('ol.Observable');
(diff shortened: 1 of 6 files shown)


src/abstractsynchronizer.js
```diff
@@ -247,13 +246,17 @@ olcs.AbstractSynchronizer.prototype.listenForGroupChanges_ = function(group) {
...
+  }).bind(this);
 
   listenAddRemove();
 
   listenKeyArray.push(group.on('change:layers', function(e) {
     contentKeys.forEach(function(el) {
-      goog.array.remove(listenKeyArray, el);
+      var i = listenKeyArray.indexOf(el);
+      goog.asserts.assert(i >= 0);
+      if (i >= 0) {
```
elemoine: Strange that you use both an assertion and a `if`.

gberaudo: The reason is that `splice()` accepts negative argument, which is definitely not what we want.
Having both the assertion and the `if` allows to always have a correct behaviour and both catch the issue at debug time.

elemoine: I don't get it. Why isn't the assertion sufficient?

gberaudo: Assertions only work in debug mode.

gberaudo: I removed the assertion. @elemoine, is this PR fine with you know?

elemoine: I didn't suggest to remove the assertion. But we've spent too much time on this already :) Please merge.

gberaudo: For the record, by removing the assertion the code matches the one in `goog.array.remove()`.

gberaudo:

PR #248, opened 2015-09-23
Prevent adding back goog library dependencies

gberaudo:


gberaudo:
Any review?

elemoine:
Cool. We could add something like this to ol3. Please merge.

gberaudo:

PR #251, opened 2015-09-24
Fix push and array handling in synchronizers

gberaudo:


fredj:
+1

gberaudo:

PR #253, opened 2015-10-01
Calculate Canvas size on initialisation

gjn:
When initialising ol3-cesium in 2D mode, the size of the canvas used is not correct. This will result in the wrong distance being calculated for the cesium camera.

The distance is calculated in the [`readView`](https://github.com/openlayers/ol3-cesium/blob/b06886cdb0958a75ff209782ac1af05bfebd6beb/src/camera.js#L378) function of the olcs.camera which is using the canvas dimensions.

### DIFF
--- a/src/ol3cesium.js
+++ b/src/ol3cesium.js
@@ -126,6 +126,9 @@ olcs.OLCesium = function(options) {
         new olcs.VectorSynchronizer(this.map_, this.scene_)
       ];
 
+  // Assures correct canvas size after initialisation
+  this.handleResize_();
+
   for (var i = synchronizers.length - 1; i >= 0; --i) {
     synchronizers[i].synchronize();
   }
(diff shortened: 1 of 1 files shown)


src/ol3cesium.js
```diff
@@ -140,6 +140,9 @@ olcs.OLCesium = function(options) {
 
   this.camera_.readFromView();
 
+  // Assures correct canvas size after initialisation
+  this.handleResize_();
```
gberaudo: Great catch @gjn.
Please move the block to line 128 (just before the synchronizer loop).
Could you also remove line 141 `this.camera_.readFromView();` which is not necessary?

gjn:
@gberaudo Updated as requested.

gberaudo:
Thank you @gjn for this patch.

PR #255, opened 2015-10-02
Update terrain provider url

### LINKED ISSUE #254: We need to update terrain provider url
Since it is now a 404, see: https://github.com/AnalyticalGraphicsInc/cesium/pull/2832

bartvde:
closes #254


gberaudo:
Right, thanks for the PR.

PR #257, opened 2015-10-07
Auto stop render loop

gberaudo:
This PR is an adaptation of [some code from Kevin Ring (NICTA)](https://github.com/TerriaJS/terriajs/blob/ebd382a8278a817fce316730d9e459bbb9b829e9/lib/Models/Cesium.js).
It allows automatically rendering the Cesium globe only when necessary, saving a lot of resource and battery on mobile devices.

The source project uses the Apache v2 license so I kept the code in a single file: `src/autorenderloop.js
` and added a comment at the top of the file. Is it OK to handle Apache2 code this way?
@kring, does it look fine for you?

This is `experimental` and `opt-in`: see the vector example for an example of use.
In the long run, it would be nice to have this directly in Cesium; there is a reflexion about it here:
https://github.com/AnalyticalGraphicsInc/cesium/issues/1865.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,8 @@
     `olcs.RasterSynchronizer.prototype.convertLayerToCesiumImageries`.
 
 * Changes
+  * Add the experimental method `olcs.OLCesium.enableAutoRenderLoop` to stop
+    rendering the globe when idle. This is based on work from Kevin Ring.
   * Port to Cesium 1.14.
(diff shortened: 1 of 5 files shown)


kring:
@gberaudo I'm happy you found the code useful!  You're using it in an acceptable way license-wise as far as I can see.  TerriaJS and Cesium both use the Apache 2.0 license so there shouldn't be any extra surprises.

gberaudo:
Any review?

gberaudo:
If there is no objection, I will merge this PR today.

PR #258, opened 2015-10-08
Verify the feature is still displayed on the map before creating the …

oterral:
…billboard

@gberaudo if you have a better idea (go through all the features of the source) to test if the feature is still displayed on the map, I take it.

### DIFF
--- a/src/featureconverter.js
+++ b/src/featureconverter.js
@@ -490,7 +490,15 @@ olcs.FeatureConverter.prototype.olPointGeometryToCesium =
     // Cesium requires the image to be loaded
     var listener = function() {
       if (!billboards.isDestroyed()) {
-        reallyCreateBillboard();
+        // We create the billboard only if the feature is still displayed on the
+        // map.
+        var features = layer.getSource().getFeatures();
+        for (var i = 0, ii = features.length; i < ii; i++) {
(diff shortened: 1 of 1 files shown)


gberaudo:
Hi @oterral, I would prefer to avoid the graph traversal in rbush + creation of an array + iteration over all features.
We could listen on the source for clear and feature removal event and cancel the creation of the billboard.
I had a try, see https://github.com/openlayers/ol3/pull/4243 and https://github.com/oterral/ol3-cesium/compare/loadbill...gberaudo:loadbill.

oterral:
@gberaudo yes you 've a much better solution. I test it.

oterral:
@gberaudo [here](https://github.com/oterral/ol3-cesium/compare/loadbill...gberaudo:loadbill#diff-055a625e546d9fc23902a2db70d609f1R535) the clear event doesn't exist on `layer`, shouldn't we listen on `source.on['removefeature', 'clear' ... `  ?

oterral:
Listening the source events it works as expected.

gberaudo:
Right @oterral, I meant to listen on the source.

oterral:
@gberaudo will you make a PR with your branch ?

gberaudo:
Thank you @oterral for spotting the issue. It is really nice to have this fixed.

Issue #245, opened 2015-09-22
olPointGeometryToCesium doesn't handle case where getImage returns null


bartvde:
See: https://github.com/openlayers/ol3-cesium/blob/7fa3360e7b063b8f04dd0f057506e7c01aa6e353/src/featureconverter.js#L429:L435

```
Uncaught TypeError: Cannot read property 'getImage' of null
```

gberaudo:
Hi Bart, what do you have in your style?

bartvde:
This is what the style object looks like before it goes into the ol3-cesium function:

![screen shot 2015-09-24 at 19 32 26](https://cloud.githubusercontent.com/assets/319678/10080966/0ac06c90-62f3-11e5-8ee2-5e563ef6f8ef.png)

PR #259, opened 2015-10-09
Allow a point geometry to have only a text style

oterral:
ol3-cesium failed to display a point feature which has only a text style, this PR fixes it.


bartvde:
this probably fixes #245 as well

gberaudo:
The PR makes sense, thank you @oterral.

PR #260, opened 2015-10-12
Use the same default text align value as in ol3

oterral:


gberaudo:
Looks good.

gberaudo:
Thank you @oterral.

PR #262, opened 2015-10-19
Fix typo 'ineartia' to 'inertia'.

poppycocker:
Fixed a typo.

### DIFF
--- a/Cesium.externs.js
+++ b/Cesium.externs.js
@@ -2543,7 +2543,7 @@ Cesium.ScreenSpaceCameraController.prototype.inertiaSpin;
 /**
  * @type {number}
  */
-Cesium.ScreenSpaceCameraController.prototype.ineartiaTranslate;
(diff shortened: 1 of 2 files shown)


fredj:
Good catch!
Can you please also change the [externs](https://github.com/openlayers/ol3-cesium/blob/master/Cesium.externs.js#L2546)?

poppycocker:
Oh, I overlooked that.
Updated a little while ago, please check it out.

gberaudo:
Hi @poppycocker, thank you for the PR.
Could you please merge the two commits together?
You may do this with:
```
git reset HEAD^
git add Cesium.externs.js
git commit --amend
```

poppycocker:
Okay. I typed commands above, and did some merges to push...
Would this be OK?

gberaudo:
After the above commands you should have force pushed to your branch.
I merged your code in #263; thank you for the PR.

poppycocker:
That makes sense then... I will study more about Git.
Thank you for helping me!!

PR #264, opened 2015-10-19
Prevent adding removed feature

gberaudo:
Verify the feature is still displayed on the map before creating the billboard.


gberaudo:
Closes #258.

PR #265, opened 2015-10-19
Remove the group from layerMap list when it is removed

oterral:
Without removing the layerGroup from the `layerMap` list, we can't re-add it.


gberaudo:
Thank you for the PR @oterral.

PR #266, opened 2015-10-19
Keep the good order if no zIndex defined

oterral:
Fixes https://github.com/geoadmin/mf-geoadmin3/issues/2712

This PR adds the groups to the list of `flattenLayers`, this keep the good order if no z-index are defined or if z-index are defined on a layer group.

The content `layerMap` array [here](https://github.com/openlayers/ol3-cesium/blob/ce86ef6ac25195c638c5ef0c7ffcf4dd2d51d51f/src/rastersynchronizer.js#L160) will decide which layers must be added to the map.

### DIFF
--- a/src/abstractsynchronizer.js
+++ b/src/abstractsynchronizer.js
@@ -89,6 +89,10 @@ olcs.AbstractSynchronizer.prototype.synchronize = function() {
  */
 olcs.AbstractSynchronizer.flattenLayers =
     function(layer, foundLayers, foundGroups, opt_zIndices) {
+  foundLayers.push(layer);
+  if (opt_zIndices) {
+    opt_zIndices[goog.getUid(layer)] = layer.getZIndex();
+  }
   if (layer instanceof ol.layer.Group) {
     foundGroups.push(layer);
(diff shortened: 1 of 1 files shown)


src/abstractsynchronizer.js
```diff
@@ -89,6 +89,10 @@ olcs.AbstractSynchronizer.prototype.synchronize = function() {
  */
 olcs.AbstractSynchronizer.flattenLayers =
     function(layer, foundLayers, foundGroups, opt_zIndices) {
+  foundLayers.push(layer);
```
gberaudo: Closure should have rejected pushing `ol.layer.Base` to an array of `ol.layer.Layer`.

gberaudo:
It is not the way the flatten function is supposed to be used.
With this PR, the groups are pushed both to the groups and layers arrays which is not what we want (see olcs.AbstractSynchronizer.prototype.removeLayer_).

So I created a PR which implement your changes but get rides of `olcs.AbstractSynchronizer.flattenLayers`.
Is it fine with you @oterral?

oterral:
Fine for me @gberaudo thanks

PR #267, opened 2015-10-19
Handle Z-Index for groups

gberaudo:
See #266 from @oterral.

This version simplifies layers traversal by getting ride of `olcs.AbstractSynchronizer.flattenLayer`.

### DIFF
--- a/src/abstractsynchronizer.js
+++ b/src/abstractsynchronizer.js
@@ -79,35 +79,6 @@ olcs.AbstractSynchronizer.prototype.synchronize = function() {
 
 
 /**
- * Populate the foundLayers and foundGroups using a breadth-first algorithm.
- * A map of layer uid to z-index is also populated.
- * @param {ol.layer.Base} layer
- * @param {Array.<ol.layer.Layer>} foundLayers Found leaves.
- * @param {Array.<ol.layer.Group>} foundGroups Found nodes.
- * @param {Object.<number, number>=} opt_zIndices Map of layer uid to z-index.
- * @protected
- */
-olcs.AbstractSynchronizer.flattenLayers =
-    function(layer, foundLayers, foundGroups, opt_zIndices) {
-  if (layer instanceof ol.layer.Group) {
-    foundGroups.push(layer);
-    var sublayers = layer.getLayers();
-    if (goog.isDef(sublayers)) {
-      sublayers.forEach(function(el) {
-        olcs.AbstractSynchronizer.flattenLayers(el, foundLayers, foundGroups,
-            opt_zIndices);
-      });
-    }
-  } else {
-    foundLayers.push(layer);
-    if (opt_zIndices) {
-      opt_zIndices[goog.getUid(layer)] = layer.getZIndex();
-    }
-  }
-};
-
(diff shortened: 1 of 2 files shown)


gberaudo:
Closes #266.

gberaudo:
Thank you again @oterral.

bartvde:
seems there is still an issue with layer groups and z-indexes:

This works fine:

```
var layers = [
  new ol.layer.Tile({
    source: new ol.source.MapQuest({layer: 'osm'})
  }),
  new ol.layer.Tile({
    source: new ol.source.TileWMS({
      url: 'http://localhost:8080/geoserver/wms',
      params: {LAYERS: 'usa:states'}
    })
  })
] ;
```
screen shot 2016-04-04 at 17 14 06

However this does not work:

```
var layers = [
  new ol.layer.Group({
    layers: [
      new ol.layer.Tile({
        source: new ol.source.MapQuest({layer: 'osm'})
      })
    ]
  }),
  new ol.layer.Tile({
    source: new ol.source.TileWMS({
      url: 'http://localhost:8080/geoserver/wms',
      params: {LAYERS: 'usa:states'}
    })
  })
];
```

screen shot 2016-04-04 at 17 15 42

bartvde:
@gberaudo shouldn't the code recurse over the group layers instead?

gberaudo:
The recursion is handled in the `orderLayers` method of `rastersynchronizer.js`.
As you can see we descend into the groups as soon as we encounter one.

I had a quick try with your example and it seems the Cesium imageries are correctly sorted in the same order as the OL3 layers but for some reason the imageries does not show up as expected in Cesium.

bartvde:
Right just saw the recursion is handled properly while running in the debugger. Ah okay, so it might be a Cesium issue instead?

gberaudo:
Maybe a bad initialization in Cesium.

bartvde:
No the sort is actually wrong, TileWMS becomes before MapQuest in the layers array. If I add explicit zIndex (0 for MapQuest group layer, 1 for TileWMS layer) the order is okay and it shows up ok.

bartvde:
Yeah right so the child of the group layer (MapQuest) gets pushed to fifo after the TileWMS layer

bartvde:
This works only wondering if there is a better way:

```
diff --git a/src/rastersynchronizer.js b/src/rastersynchronizer.js
index f4cf72f..62d5858 100644
--- a/src/rastersynchronizer.js
+++ b/src/rastersynchronizer.js
@@ -152,11 +152,16 @@ olcs.RasterSynchronizer.prototype.orderLayers = function() {
     layers.push(olLayer);
     zIndices[goog.getUid(olLayer)] = olLayer.getZIndex();
 
+    var len = fifo.length;
     if (olLayer instanceof ol.layer.Group) {
       var sublayers = olLayer.getLayers();
       if (goog.isDef(sublayers)) {
         sublayers.forEach(function(el) {
-          fifo.push(el);
+          if (len === 0) {
+            fifo.push(el);
+          } else {
+            fifo.unshift(el);
+          }
         });
       }
     }
```

PR #268, opened 2015-10-20
Port to OL 3.10.1 and Cesium 1.14

gberaudo:
Remove saturation, brightness and hue support which were removed in OL 3.


gberaudo:
Merging for preparing a new release.

PR #269, opened 2015-10-21
Initialize the root primitive visibility according to the layer

oterral:


gberaudo:
Thank you for the fix @oterral.

PR #271, opened 2015-10-22
Prepare release 1.9

gberaudo:
Update changelog.


gberaudo:
Fixed vector example where render loop was not always restarted.

gberaudo:
Merging

PR #274, opened 2015-10-23
Add epsilon to camera change check

gberaudo:
Due to inertia and computation instabilities, synchronization may be very long.
This change allows faster synchronization.


fredj:
LGTM

gberaudo:

PR #279, opened 2015-11-02
Fix sky rendering

oterral:
Using `time` the sky was not rendered anymore.

### DIFF
--- a/src/ol3cesium.js
+++ b/src/ol3cesium.js
@@ -156,10 +156,11 @@ olcs.OLCesium = function(options) {
 
   this.cesiumRenderingDelay_ = new goog.async.AnimationDelay(function(time) {
     if (!this.blockCesiumRendering_) {
+      var julianDate = Cesium.JulianDate.now();
       this.scene_.initializeFrame();
       this.handleResize_();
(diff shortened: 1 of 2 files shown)


src/ol3cesium.js
```diff
@@ -159,7 +159,7 @@ olcs.OLCesium = function(options) {
       this.scene_.initializeFrame();
       this.handleResize_();
       this.dataSourceDisplay_.update(time);
```
gberaudo: The time passed to dataSourceDisplay is also incorrect.
We should create a julianDate and pass it to both the dataSourceDisplay and the scene.
```
var julianDate = Cesium.JulianDate.now();
this.dataSourceDisplay_.update(julianDate);
this.scene_.render(julianDate);
```

This should be added to Cesium.externs.js
```
/**
 * @return {Cesium.JulianDate}
 */
Cesium.JulianDate.now = function() {};
```

oterral:
PR updated  @gberaudo

fredj:
Thanks @oterral

gberaudo:
Tanks @oterral and @fredj.

PR #281, opened 2015-11-03
Honour pixel ratio and allow degrading quality for performance

gberaudo:
Add `olcs.OLCesium.setResolutionScale` to allow improving performance at the cost of quality.
Automatically use device pixel ratio to configure the Webgl 3D globe.


fredj:
LGTM

gberaudo:
Not yet ready, there is an issue with view synchronization. I am looking at it.

gberaudo:
Should be fixed now.

PR #285, opened 2015-11-09
Test if the source of a ol.layer.Image is usable by the vector synchr…

oterral:
…onizer.

This PR avoid an assert error when a `ol.layer.Image` use a non-`ol.source.ImageVector` like `ol.source.WMS` for example.


gberaudo:
Thanks @oterral.

PR #287, opened 2015-11-16
Fix group removal

gberaudo:
When a layer group is removed, first try to remove the group
counterparts. If unsucessful, try to remove each of the group children.


fredj:
LGTM

gberaudo:
Thanks for the review @fredj.

PR #289, opened 2015-11-19
Compile OL3-Cesium together with OL3

gberaudo:
OL3-Cesium used to be compiled against OL3 externs.
This was unnatural for a Closure project and had shortcomings:
- a custom closure compiler was required;
- we were restricted to OL3 @api symbols;
- OL3 externs were sometimes buggy;
- the goog library was included twice;
...

This change bundles the OL3 code together with OL3-Cesium, which is already what we do in Ngeo.
This will allow adding client side raster reprojection in the future.

### DIFF
--- a/closure-util.json
+++ b/closure-util.json
@@ -1,4 +1,4 @@
 {
-  "compiler_url": "http://dev.camptocamp.com/files/closure-compiler/compiler-20150903.zip",
-  "library_url": "https://github.com/google/closure-library/archive/b129bb.zip"
+  "Allowed keys, see":  "https://github.com/openlayers/closure-util/blob/master/default-config.json",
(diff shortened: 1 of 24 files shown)


closure-util.json
```diff
@@ -1,4 +1,4 @@
 {
```
fredj: What about removing this file?

gberaudo:
@fredj, I removed the file and instead put instructions in the `CONTRIBUTING.md`
file. I also fixed travis.

fredj:
+1

gberaudo:
Thanks for the review @fredj.
We decided on this at FOSDEM 2015 Seoul; I will now merge it.

PR #290, opened 2015-11-24
Honor devicePixelRatio like in Cesium

gberaudo:
The Cesium team updated their default handling of pixel ratio.
On devices with pixelRatio > 1, the buffer used for rendering the scene will be smaller than the size of the canvas (clientWidth, clientHeight) and scaled afterwards.
This change drastically improves FPS on smartphones and reduces bandwidth usage.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@
   * OL3-Cesium is now compiled together with OL3. A custom closure compiler
     build is no more required.
 
+* Changes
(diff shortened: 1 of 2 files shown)


gberaudo:
See https://github.com/AnalyticalGraphicsInc/cesium/pull/3233

fredj:
LGTM

gberaudo:
Thanks for the review @fredj.

PR #292, opened 2015-11-27
Revert "Honor devicePixelRatio like in Cesium"

gberaudo:
Reverts openlayers/ol3-cesium#290

The Cesium team is thinking about reverting their change:
https://github.com/AnalyticalGraphicsInc/cesium/issues/3249.


fredj:
+1

gberaudo:
Thanks for the review @fredj. Merging it now so that we can have a release today.

PR #297, opened 2015-12-10
Take into account scale on icons

bartvde:
otherwise this might happen: 
screen shot 2015-12-10 at 13 19 50


gberaudo:
Thanks @bartvde for the PR.
Please add `scale` to the externs in Cesium.externs.js#Cesium.optionsBillboardCollectionAdd.

bartvde:
Ah there is an externs for the cesium properties, didn't know that, will update

gberaudo:
Great thanks.

PR #305, opened 2016-01-13
Fix layer group removal

gberaudo:
Layers in a layergroup were not properly removed.

Remove counterparts associated to individual layers when there is no
counterpart associated to the group itself.


fredj:
LGTM

gberaudo:
Thanks for the review @fredj.

PR #307, opened 2016-01-20
Update bing key

gberaudo:
Update the OpenLayers Bing key for the online examples:
http://openlayers.org/ol3-cesium/examples/

### DIFF
--- a/examples/rastersync.js
+++ b/examples/rastersync.js
@@ -41,7 +41,7 @@ ol3d.setEnabled(true);
 var addBingMaps = function() {
   ol2d.addLayer(new ol.layer.Tile({
     source: new ol.source.BingMaps({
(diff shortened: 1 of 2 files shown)


gberaudo:
See https://github.com/openlayers/ol3/pull/4691

ahocevar:
Please don't use the OpenLayers API key. We need to be very restrictive with the use of that key, to avoid that we go over quota.

tschaub:
I agree it would be best to restrict this key to use in the OpenLayers examples only.  Keys are freely available at https://www.bingmapsportal.com/, so I'd suggest registering for a dedicated ol3-cesium one.

gberaudo:
Thanks for the feedbacks @ahocevar and @tschaub.

PR #311, opened 2016-01-20
Use OSM instead of Bing

gberaudo:
The Bing key was expired.
Add lighting effect.


ahocevar:
+1, please merge.

gberaudo:

PR #322, opened 2016-02-29
Accept a mode=dev parameter in examples

gberaudo:
With this parameter, unminified source code is loaded instead of minified one. This is helpful for debugging.


fredj:
LGTM

gberaudo:
Thanks for the review @fredj.

PR #330, opened 2016-03-24
Fix test to createShadowVolume property

gberaudo:
Should fix https://github.com/openlayers/ol3-cesium/issues/329.


fredj:
LGTM

oterral:
when cesium is compiled, isn't `createShadowVolume` changed  to a smaller name?

oterral:
Thx for the quick fix.

gberaudo:
Thanks for the review.

PR #332, opened 2016-03-31
Set release date

gberaudo:


gberaudo:
Merging trivial change for release.

PR #337, opened 2016-04-04
Fix layer ordering

gberaudo:
Layer group sublayers were appended to the queue instead of being
prepended. As a consequence, a single layer positioned after a group
was ordered first.

This commit prepends the sublayers so that they are handled first.


gberaudo:
See https://github.com/openlayers/ol3-cesium/pull/267#issuecomment-205343709.
@bartvde, could you please confirm it fixes your issue?

bartvde:
thanks @gberaudo this works great, code looks sane as well. LGTM. Thanks for fixing so quickly!

gberaudo:
Thanks for reporting the issue @bartvde.

PR #340, opened 2016-04-08
Fix frustum aspect ratio and check earlier

gberaudo:
This PR fixes a nasty bug and adds early checks.

### DIFF
--- a/src/ol3cesium.js
+++ b/src/ol3cesium.js
@@ -306,7 +306,7 @@ olcs.OLCesium.prototype.getEnabled = function() {
  * @api
  */
 olcs.OLCesium.prototype.setEnabled = function(enable) {
-  if (this.enabled_ == enable) {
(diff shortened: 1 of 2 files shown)


src/ol3cesium.js
```diff
@@ -436,3 +443,18 @@ olcs.OLCesium.prototype.setResolutionScale = function(value) {
...
   }
 };
+
+
+/**
+ * Check if OL3 map is not properly initialized.
+ * @private
+ */
+olcs.OLCesium.prototype.throwOnUnitializedMap_ = function() {
+  var map = this.map_;
```
fredj: `ol.View#isDef()` can be used here

gberaudo: Thanks, using it now.

gberaudo:
Thanks for the review @fredj. If there is no other comments I will merge this PR tomorrow.

PR #342, opened 2016-04-11
Enabled support for arrays in attributions (https://github.com/openlayers/ol3-cesium/issues/341

jmgomezpoveda:


gberaudo:
`el.getHTML()` returns a string. It can not be iterated like an array.

jmgomezpoveda:
That is weird, since the problem is that sometimes el.getHTML() is returning an Array (hence the if above). Maybe getHTML() should not return an array, and the actual cause has to be looked into getHTML().

gberaudo:
The documentation says it should be a string: http://openlayers.org/en/master/apidoc/ol.Attribution.html

jmgomezpoveda:
Note the different syntax used in that sample:
```
  attributions: [
    new ol.Attribution({
      html: 'All maps © ' +
          '<a href="http://www.opencyclemap.org/">OpenCycleMap</a>'
    }),
    ol.source.OSM.ATTRIBUTION
  ],
```

As in the negative sample I provided, where the html does contain an array:
`attributions: [new ol.Attribution({ html: ['© xxx'] })]`

I can easily avoid this in the custom attributions I create, but it turns out that attributions created by some default classes (OSM and/or MapTiles) also create arrays.

jmgomezpoveda:
Actually, my bad. All default classes create attributions without arrays. It was only me in custom attributions (whose syntax I got from some samples around, and which work fine in vanilla ol3) who was causing them to be created as arrays, using the syntax above. So, I can fix the issue in my code without any change in openlayers or ol3-cesium.

Now, I wonder if ol3 should somehow validate for that syntax not to be used, or if ol3-cesium should try to tolerate it (or just prevent it by raising an error as in https://github.com/openlayers/ol3-cesium/issues/341).

gberaudo:
Adding code for every possible user mistake would make the libraries very huge.
So I think it is better to keep think as-is.

jmgomezpoveda:
The same syntax is used in the CartoDB sample code in https://cartodb.com/basemaps/

gberaudo:
That is an error of CartoDB. Feel free to notify them of the error.
Actually calling the ol.Attribution constructior is now optionnal. Syntaxes like these are valid:
```
attributions: 'some attribution'
```
```
attributions: ['first attribution', 'second attribution']
```

jmgomezpoveda:
Thanks. I have just notified CartoDB.

Trying these alternative syntaxes (without the new ol.Attribution...) I get the following error:
> TypeError: sourceAttribution.intersectsAnyTileRange is not a function

PR #356, opened 2016-05-10
Add support for Cesium rectangles

jmgomezpoveda:
This PR is to support Cesium rectangles.
Closes: https://github.com/openlayers/ol3-cesium/issues/355.

We should document this functionality somewhere, as it is not straightforward.

`geometry.set('olcs.polygon_kind', 'rectangle');`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,10 @@
 # Changelog
 
 * Changes
-  * Added support for Cesium rectangles via the olcs.polygon_kind property.
+  * Add support for drawing rectangles according to the longitude and latitude
+    curves instead of straight lines. This functionality can be activated by
+    setting the olcs.polygon_kind property to 'rectangle' on the OpenLayers
+    geometry.
 
 ## v 1.15 - 2016-04-28
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 # Changelog
 
+* Changes
+  * Added support for Cesium rectangles via the olcs.polygon_kind property.
+
 ## v 1.15 - 2016-04-28
 
 * Changes
--- a/Cesium.externs.js
+++ b/Cesium.externs.js
@@ -1483,6 +1483,14 @@ Cesium.PolygonOutlineGeometry = function(object) {};
 
 
 /**
+ * @constructor
+ * @param {Object=} object
+ * @extends {Cesium.Geometry}
+ */
+Cesium.RectangleOutlineGeometry = function(object) {};
+
+
+/**
  * @typedef {{
  * positions: !Array.<Cesium.Cartesian3>,
  * vertexFormat: number
@@ -1497,6 +1505,23 @@ Cesium.optionsPolylineGeometry;
  * @extends {Cesium.Geometry}
  */
 Cesium.PolylineGeometry = function(object) {};
+
+
+/**
+ * @typedef {{
+ * rectangle: {Cesium.Rectangle},
+ * vertexFormat: number
+ * }}
+ */
+Cesium.optionsRectangleGeometry;
+
+
+/**
+ * @constructor
+ * @param {Object=} object
+ * @extends {Cesium.Geometry}
+ */
+Cesium.RectangleGeometry = function(object) {};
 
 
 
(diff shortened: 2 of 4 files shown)


Cesium.externs.js
```diff
@@ -1483,6 +1483,14 @@ Cesium.PolygonOutlineGeometry = function(object) {};
 
 
 /**
+ * @constructor
+ * @param {Object=} object
```
gberaudo: A typedef should be created for this parameter based on the options in https://cesiumjs.org/Cesium/Build/Documentation/RectangleOutlineGeometry.html?classFilter=outline#RectangleOutlineGeometry.

Cesium.externs.js
```diff
@@ -1499,6 +1507,23 @@ Cesium.optionsPolylineGeometry;
...
+ * rectangle: {Cesium.Rectangle},
+ * vertexFormat: number
+ * }}
+ */
+Cesium.optionsRectangleGeometry;
+
+
+/**
+ * @constructor
+ * @param {Object=} object
```
gberaudo: Hmm, I see you actually created the typedef but did not used it. It should be used instead of `object`.
I checked the code of the extern and there are other places like this; these are very weird errors and need a fix.

gberaudo:
The vector example could be updated, changing one of the existing rectangle to use the `olcs.polygon_kind` `rectangle` property.

A note should be added to the CHANGES.md.

You are also required to sign the OpenLayers CLA at https://github.com/openlayers/cla.

Great change, I am looking forward to merge this PR.

jmgomezpoveda:
Hi, thanks for the review and the feedback:
* Completed and used the rectangle typedefs
* Updated the vector example with an additional polygon with the property set.
* Added a note to the CHANGES.md.
* Signed the OpenLayers CLA

jmgomezpoveda:
Also, I have enabled support for the height property if vertices have a height. Cesium rectangles do not support per position heights, so I have calculated the average height of the vertices. I could just have used the height of the first vertex, assuming they would all be the same, but this seems a bit more generic and accurate.

gberaudo:
I think it is better to take the maximum height of the vertices to make sure the corner of the rectangle are visible.

CHANGES.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+* Changes
+  * Added support for Cesium rectangles via the olcs.polygon_kind property.
```
gberaudo: The description could be more detailed: `Add support for drawing rectangles according to the longitude and latitude curves instead of straight lines. This functionality can be activated by setting the olcs.polygon_kind property to 'rectangle' on the OpenLayers geometry.`

src/featureconverter.js
```diff
@@ -397,38 +397,69 @@ olcs.FeatureConverter.prototype.olPolygonGeometryToCesium =
...
-    if (i == 0) {
-      hierarchy.positions = positions;
-    } else {
-      hierarchy.holes = {
-        // always update Cesium externs before adding a property
-        positions: positions
-      };
-      hierarchy = hierarchy.holes;
+  var fillGeometry, outlineGeometry;
+  if (feature.getGeometry().get('olcs.polygon_kind') === 'rectangle') {
```
gberaudo: You should only enter here if there are exactly 4 elements in the coordinates array.

jmgomezpoveda: Whereas that check makes sense, polygons in OpenLayers have to be closed, repeating the first vertex at the end. So, there should be 5 coordinates.

jmgomezpoveda: On the height, average vs. maximum, a rectangle only has 4 points (plus the closing one), so they are all corners. Anyway, I have changed it as suggested.

gberaudo:
Your name does not appear on https://github.com/openlayers/cla/blob/master/signed-agreements.md. Have you submitted the individual contributor agreement?

jmgomezpoveda:
Yes, via the Google form that you have me.

gberaudo:
OK, looks like the file is not updated automatically.
Thanks for your contribution, could you please squash the commits and then I will merge?

gberaudo:
Thanks again @jmgomezpoveda, I just merged your PR.

jmgomezpoveda:
Thanks for guiding me through the process. Looking forward to making more contributions in the future.

PR #358, opened 2016-05-13
Configure proxy to use to retrieve tiles.

fxprunayre:
Add support to set a proxy for Cesium to load layer not accessible due to missing CORS headers (eg. when user can't modify configuration of the mapservice used). This functionality can be activated by setting the olcs.proxy property to the OpenLayers source.

The following error occurs if CORS is not configured on the remote service:
```
Image from origin 'http://remote.service.org' has been blocked from loading 
by Cross-Origin Resource Sharing policy: 
No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

The ```olcs.proxy``` property can contains:
* a function
```
source.set('olcs.proxy', function (url) {
    return '/myproxy?url=' + encodeURIComponent(url);
});
```
* a string pointing to the proxy. In that case a ```Cesium.DefaultProxy``` will be configured on this URL (see https://cesiumjs.org/Cesium/Build/Documentation/DefaultProxy.html).
```
source.set('olcs.proxy', '/myproxy/url');
```

Related discussions:
* http://stackoverflow.com/questions/30738371/proxy-for-wms-layers

### DIFF
--- a/src/core/olimageryprovider.js
+++ b/src/core/olimageryprovider.js
@@ -40,6 +40,17 @@ olcs.core.OLImageryProvider = function(source, opt_fallbackProj) {
 
   this.ready_ = false;
 
+  var proxy = this.source_.get('olcs.proxy');
+  if (goog.isDef(proxy)) {
+    if (goog.isFunction(proxy)) {
+      this.proxy_ = {
+        'getURL': proxy
+      };
+    } else if (typeof proxy === 'string') {
+      this.proxy_ = new Cesium.DefaultProxy(proxy);
+    }
+  }
+
   this.errorEvent_ = new Cesium.Event();
 
   this.emptyCanvas_ = document.createElement('canvas');
@@ -120,7 +131,8 @@ Object.defineProperties(olcs.core.OLImageryProvider.prototype, {
   },
 
   proxy: {
-    get: function() {return undefined;}
+    get: /** @this {olcs.core.OLImageryProvider} */
+        function() {return this.proxy_;}
   },
 
   hasAlphaChannel: {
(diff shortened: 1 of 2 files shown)


src/core/olimageryprovider.js
```diff
@@ -40,6 +40,17 @@ olcs.core.OLImageryProvider = function(source, opt_fallbackProj) {
 
   this.ready_ = false;
 
+  var proxy = this.source_.get('olcs.proxy');
+  if (goog.isDef(proxy)) {
+    if (goog.isFunction(proxy)) {
+      this.proxy_ = {
+        getURL: proxy
```
gberaudo: I think you need to quote the key: `'getURL'` to make sure the name is not minified by closure in advanced mode.

gberaudo:
I added a small comment, otherwise it looks good.
Have you signed the OpenLayers contributor aggreement at https://github.com/openlayers/cla/blob/master/signed-agreements.md?

fxprunayre:
> Have you signed the OpenLayers contributor aggreement

Yep.

gberaudo:
OK, looks like the file is not updated automatically.
I will first merge the PR from @jmgomezpoveda then you will have to rebase and squash the commits.
Thanks for your contribution.

gberaudo:
OK, https://github.com/openlayers/ol3-cesium/pull/356 is now merged.

gberaudo:
I missed your rebase. Thanks again for your contribution.

fxprunayre:
Perfect. Thanks.

amacar:
Can we get any example? 
I tried using this setting but all I get is black screen. If I later disable this layer from layerswitcher I get baselayer shown normally.

PR #369, opened 2016-06-09
Support arbitrary line widths on windows

gberaudo:
All browsers on Windows only support a maximum WebGL width of 1.
This commit adds support for arbitrary width.


gberaudo:
I tested on Windows and other plateforms and it looks good. Merging this.

PR #370, opened 2016-06-15
Use fs-extra as fs and remove usage of graceful-fs

fredj:
See https://github.com/openlayers/ol3/pull/5151


gberaudo:
Cool, thanks.

PR #385, opened 2016-08-29
Port to OpenLayers 3.18.1

gberaudo:


ahocevar:
Don't you want to wait for OpenLayers v3.18.1? It will have fixes for at least two regressions that were reported.

gberaudo:
Thanks @ahocevar for notifying me. Do we already have an ETA?
Cesium will come on Thursday; I can certainly wait until next week for the portings.

ahocevar:
That should be fine. One of the two fixes is already merged, the other is awaiting review. So I'm expecting OpenLayers 3.18.1 to be out today or tomorrow.

gberaudo:
Great! thanks.

gberaudo:
Updated with 3.18.1. See https://github.com/openlayers/ol3/pull/5803.

gberaudo:
OK, I think this PR makes consensus. Merging it now.
Thanks @ahocevar for being so fast publishing the OL3 patch version.

PR #387, opened 2016-09-02
Make sure dist is present in package published on npm

gberaudo:
See https://www.npmjs.com/package/ol3-cesium.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,2 @@
+.build
+examples/example-list.js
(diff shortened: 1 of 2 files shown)


ahocevar:
The change does not do what the title suggests. Wrong branch maybe?

gberaudo:
No it is correct. The .npmignore takes precedence over the .gitignore.

ahocevar:
Ah ok. Now I see what you were trying to do. Alright! 👍

gberaudo:

PR #391, opened 2016-09-08
Add DataSources externs

oterral:


Cesium.externs.js
```diff
@@ -2585,6 +2585,56 @@ Cesium.DataSource = function() {};
...
+ *   fill: Cesium.Color,
+ *   clampToGround: boolean
+ * }}
+ */
+Cesium.GeoJsonDataSourceLoadOptions;
+
+/**
+ * @param {string|Object} data
+ * @param {!Cesium.GeoJsonDataSourceLoadOptions} options
+ * @returns {Cesium.DataSource|Promise.<Cesium.DataSource>}
```
gberaudo: There should be no s in `@return`. Could you also fix the other occurences in the file?

oterral: Ok i can do that

oterral:
`returns` replaced

gberaudo:
Great, thanks.

PR #396, opened 2016-09-14
Set ol.DEBUG define to false

fredj:


gberaudo:
Thanks, please merge.

fredj:
`ol.DEBUG` was introduced after the 3.18.2 ol3 release. we need to wait for 3.19.0

gberaudo:
I see, let's wait for the next stable OL3 version.

fredj:
#407

PR #400, opened 2016-09-15
Get rid of goog.dom.getNextElementSibling

fredj:
closure-library code:
https://github.com/google/closure-library/blob/v20160822/closure/goog/dom/dom.js#L1252-L1262


gberaudo:
That seems to be handled by all modern browsers. See https://developer.mozilla.org/en-US/docs/Web/API/NonDocumentTypeChildNode/nextElementSibling.

Thanks and please merge.

PR #402, opened 2016-09-15
Use export tag instead of goog.exportProperty

fredj:
The build file (`dist/ol3cesium.js`) is the same as before.


gberaudo:
Thanks, please merge.

PR #405, opened 2016-10-07
Entity tracking

gberaudo:


src/ol3cesium.js
```diff
@@ -207,6 +224,125 @@ olcs.OLCesium = function(options) {
...
+   */
+  this.boundingSphereScratch_ = new Cesium.BoundingSphere();
+
+  var eventHelper = new Cesium.EventHelper();
+  eventHelper.add(this.scene_.postRender, olcs.OLCesium.prototype.updateTrackedEntity_, this);
+};
+
+
+Object.defineProperties(olcs.OLCesium.prototype, {
+  trackedFeature: {
```
fredj: add quotes to make sure the property is not renamed by the compiler ?

gberaudo: There are no quotes in `src/core/olimageryprovider.js` so I did it the same here.
But I agree not quoting looks strange. I will check what the compiler actually generates.

fredj: maybe the compiler is aware of `Object.defineProperties` and don't rename the keys

fredj: (and +1 for using `Object.defineProperties` !)

gberaudo:
Updated the PR and the example to use stacked views. This feature is currently limited to stacked views.

gberaudo:
@fredj, could you please check the example and merge if you think it is good?

PR #411, opened 2016-10-21
Allow setting target rendering frame rate

schmidtk:
I added a function to set the target rendering frame rate for Cesium. Our application is using this to reduce the frame rate while parsing large data sets, which speeds up parsing significantly while still providing occasional feedback to the user that data is loading.

@ahocevar or @gberaudo could you review please? Thanks!

### DIFF
--- a/src/ol3cesium.js
+++ b/src/ol3cesium.js
@@ -1,7 +1,5 @@
 goog.provide('olcs.OLCesium');
 
-goog.require('goog.async.AnimationDelay');
-goog.require('goog.async.Delay');
 goog.require('olcs.AutoRenderLoop');
 goog.require('olcs.Camera');
 goog.require('olcs.RasterSynchronizer');
@@ -198,11 +196,25 @@ olcs.OLCesium = function(options) {
   }
 
   /**
-   * Delay to render the Cesium scene.
-   * @type {goog.async.AnimationDelay|goog.async.Delay}
+   * Time of the last rendered frame, as returned by `performance.now()`.
+   * @type {number}
    * @private
    */
-  this.cesiumRenderingDelay_ = new goog.async.AnimationDelay(this.render_, undefined, this);
+  this.lastFrameTime_ = 0;
+
+  /**
+   * The identifier returned by `requestAnimationFrame`.
+   * @type {number|undefined}
+   * @private
+   */
+  this.renderId_ = undefined;
+
+  /**
+   * Target frame rate for the render loop.
+   * @type {number}
+   * @private
+   */
+  this.targetFrameRate_ = Number.POSITIVE_INFINITY;
 
   /**
    * @private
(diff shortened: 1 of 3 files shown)


schmidtk:
@ahocevar do you have time to review this? We're working on open sourcing parts of our application and are hoping to get some of our local `ol3-cesium` changes merged into the baseline. Thanks, and sorry if you're busy!

gberaudo:
Hi @schmidtk, sorry I was in vacations.

I am favorable in adding such a feature.

However I see two issues with this PR:
- it adds more dependency on `goog`: `window.requestAnimationFrame` should be used instead;
- it duplicaties some functionnality with the `setBlockCesiumRendering` method: passing a delay of `Number.POSITIVE_INFINITY` should mean no blocking and 0 should mean blocking.

schmidtk:
Thanks for the feedback! I'm replacing `goog.async` calls with `requestAnimationFrame` and modeling the animation loop to act more like the [CesiumWidget](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Widgets/CesiumWidget/CesiumWidget.js#L58) rendering loop.

The `setBlockCesiumRendering` functionality will still be somewhat duplicated by setting the target frame rate <= 0, but the former allows you to block rendering without needing to know the previous frame rate when you want to render again.

schmidtk:
@gberaudo I reworked the render loop, so `goog.async` is no longer used. This version is more in line with how the Cesium Widget's render loop works.

gberaudo:
Thanks @schmidtk, I will have a last look tomorrow before merging.

gberaudo:
Sorry I was longer than expected. It looks good. I have only found a small regression regarding warmup. Could you please have a look to it?
Would you alse be able to add a setFrameRate button (with a low hardcoded value) to the vector example (which uses auto render loop)? Then it will be easier to test and spot regressions.

schmidtk:
Sorry about that, I missed that case. Calling `setEnabled` during the warm up will prevent disabling rendering after the warm up timeout now. I removed the frame rate override/reset I had previously added during warm up rendering to avoid similar conflicts during initialization routines.

I also added an input to set the frame rate in the vectors example.

src/ol3cesium.js
```diff
@@ -566,16 +578,17 @@ olcs.OLCesium.prototype.warmUp = function(height, timeout) {
     csCamera.position = ellipsoid.cartographicToCartesian(position);
   }
 
-  // enable rendering and use the maximum frame rate while preloading
-  var targetFrameRate = this.targetFrameRate_;
-  this.targetFrameRate_ = Number.POSITIVE_INFINITY;
+  // temporarily enable rendering
   this.enabled_ = true;
+  this.warmingUp_ = true;
```
schmidtk: Fixed.

gberaudo:
Thank you for your contribution and for your reactivity to adapt the code.
Since you where first I will merge your PR now. @oterral could you please adjust your PR #416?

oterral:
Yes

schmidtk:
Thanks, @gberaudo!

PR #412, opened 2016-10-31
Update dependencies

fredj:


fredj:
@gberaudo I've updated cesium to 1.27

gberaudo:
Have you run the examples already?

gberaudo:
OK, I just tested every examples, they work. Thanks.

PR #413, opened 2016-11-02
fixed polygon holes

jbo023:
Polygons with Holes didn't work correctly, this pullrequest fixes the problem.


fredj:
Thanks @jbo023, can you please fix the eslint issues?
should be `if (!hierarchy.holes) {` instead of `if(!hierarchy.holes){`

gberaudo:
@jbo023, could you also fix the type in the Cesium externs?

gberaudo:
Thanks

PR #416, opened 2016-11-09
Use time from options

oterral:
This PR allow to specify a time for the Cesium rendering.

@gberaudo and @fredj review please ?

### DIFF
--- a/src/ol3cesium.js
+++ b/src/ol3cesium.js
@@ -28,6 +28,14 @@ olcs.OLCesium = function(options) {
   this.map_ = options.map;
 
   /**
+   * @type {!Function}
+   * @private
+   */
+  this.time_ = options.time || function() {
+    return Cesium.JulianDate.now();
+  };
+
+  /**
    * No change of the view projection.
    * @private
    */
@@ -361,7 +369,7 @@ olcs.OLCesium.prototype.onAnimationFrame_ = function(frameTime) {
   // time to render a frame, save the time
   this.lastFrameTime_ = frameTime;
 
-  var julianDate = Cesium.JulianDate.now();
+  var julianDate = this.time_();
   this.scene_.initializeFrame();
   this.handleResize_();
   this.dataSourceDisplay_.update(julianDate);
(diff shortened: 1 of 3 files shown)


schmidtk:
The time is used on each render call, so I'm curious why you want to provide a static time. The time provided with Cesium should be increasing with each frame.

This is going to conflict with my PR (#411), which is using the time from `requestAnimationFrame` to determine when to render frames. I could pass that frame time to Cesium in place of the old `Cesium.JulianDate.now()` code.

oterral:
Because I want the sun to be always at the same place. But I'm open to suggestions to achieve that.

schmidtk:
Ahh, I didn't poke into what the time provided to Cesium was controlling. Now I'm on the same page.

I refactored the render loop code a bit so we're going to have a merge conflict between our PR's, but the changes are still compatible.

src/ol3cesium.js
```diff
@@ -340,7 +346,7 @@ olcs.OLCesium.prototype.updateTrackedEntity_ = function() {
     bs.radius = 1;
   }
   this.entityView_ = new Cesium.EntityView(trackedEntity, scene, scene.mapProjection.ellipsoid);
-  this.entityView_.update(Cesium.JulianDate.now(), bs); // FIXME: have a global management of current time
+  this.entityView_.update(this.time_, bs); // FIXME: have a global management of current time
```
gberaudo: Instead of a constant time I would expect a function returning a time.
The application could then choose to:
- pass Cesium.JulianDate.now (current behaviour);
- pass a function returning a constant time (what you propose);
- pass a function driven by a UI component (similar to the time component in Cesium).

oterral: Make sense.

oterral:
Updated with a function instead of a constant

oterral:
Rebased.

src/ol3cesium.js
```diff
@@ -28,6 +28,14 @@ olcs.OLCesium = function(options) {
   this.map_ = options.map;
 
   /**
+   * @type {!Function}
+   * @private
+   */
+  this.time_ = options.time || function() {
```
gberaudo: You need to add the time parameter to the olcsx.OLCesiumOptions typedef in olcsx.js.
The type should be `function(): Cesium.JulianDate`.

gberaudo:
@oterral, would you please modify the main example to use this feature?

I imagine we could have a checkbox to select between real time and constant time.
It is fine to have shadows always enabled.

oterral:
Yes sure

oterral:
example added

gberaudo:
Thanks

PR #433, opened 2017-01-20
Update Cesium private variables in render loop

schmidtk:
Updated the tile checks in `olcs.AutoRenderLoop#postRender` to match those used in `Cesium.Camera#update`. Considered using the `_suspendTerrainAdjustment` private flag in the Cesium camera, but that seemed more fragile to change and less clear to someone trying to understand the olcs code.


schmidtk:
This fix is easily tested by turning on the render loop in any example:

`ol3d.enableAutoRenderLoop()`

Then watching the CPU usage of the tab in Chrome's task manager. Previously the tab would be spinning on the caught error, now it should drop to ~0%.

PR #437, opened 2017-02-02
Use GroundPrimitive for line

oterral:
It corresponds to the hack in Cesium.KmlDataSource.

### DIFF
--- a/src/featureconverter.js
+++ b/src/featureconverter.js
@@ -350,21 +350,37 @@ olcs.FeatureConverter.prototype.olLineStringGeometryToCesium = function(layer, f
     material: this.olStyleToCesium(feature, olStyle, true)
   });
 
-  // Handle both color and width
-  var outlineGeometry = new Cesium.PolylineGeometry({
+  var geometryOptions = {
     // always update Cesium externs before adding a property
     positions: positions,
     width: this.extractLineWidthFromOlStyle(olStyle),
     vertexFormat: appearance.vertexFormat
-  });
+  };
 
-  var outlinePrimitive = new Cesium.Primitive({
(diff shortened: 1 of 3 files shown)


src/featureconverter.js
```diff
@@ -350,21 +350,37 @@ olcs.FeatureConverter.prototype.olLineStringGeometryToCesium = function(layer, f
...
     positions: positions,
     width: this.extractLineWidthFromOlStyle(olStyle),
     vertexFormat: appearance.vertexFormat
-  });
+  };
+
+  var outlinePrimitive;
+  var heightReference = this.getHeightReference(layer, feature, olGeometry);
+
+  if (heightReference == Cesium.HeightReference.CLAMP_TO_GROUND) {
```
gberaudo: If the positions are 3D (just test the first one), could you disregard the `CLAMP_TO_GROUND` height reference?

oterral: You mean if the coordinates is x,y,z. we deactivate the clampToGround?

gberaudo: yes

oterral: We don't do this for polygons why doing it for lines ? and that doesn't make sense if the user specify clampToGround  he expects the dtat are clamped to ground whatever the z value.

gberaudo: OK, I checked the KML documentation and they do like you propose: https://developers.google.com/kml/documentation/altitudemode. So OK to merge. However, please update the README.md.

gberaudo:
Thanks @oterral.

oterral:
Thx for the review @gberaudo

PR #445, opened 2017-02-28
Rename ol3-cesium to ol-cesium

fredj:


fredj:
@gberaudo ready

gberaudo:
Thanks

PR #448, opened 2017-03-03
Port to Cesium 1.31

gberaudo:


gberaudo:
Thanks for the review @fredj.

PR #449, opened 2017-03-06
Get ride of remaining closure library usage

gberaudo:
Supersedes #366.


gberaudo:
Thanks for the review @fredj.

PR #453, opened 2017-03-31
Don't reject WMTS sources by default

fredj:


src/core.js
```diff
@@ -368,11 +367,6 @@ olcs.core.tileLayerToImageryLayer = function(olLayer, viewProj) {
   let provider = null;
 
   const source = olLayer.getSource();
-  // handle special cases before the general synchronization
-  if (source instanceof ol.source.WMTS) {
```
gberaudo: Please modify the changelog to state we now handle WMTS sources, expecting the tilegrid is the same.

fredj:
@gberaudo changelog updated

zl142900:
Does ol-cesium support WMTS now?Why didn't I achieve it?

gberaudo:
@zl142900, there is no example so I am not sure if everything is in OL-Cesium or not.
The best approach is to run your code step by step and see if there is an early exit somewhere.

zl142900:
At present, my problem is that the WMTS can only be loaded in two dimensions. If you switch to three dimensions, the WMTS cannot be loaded

Do I need to set  in core.js
 if (source instanceof olSourceWMTS) {
      //What should I write
 }

PR #455, opened 2017-05-04
Update dependencies, OL 4.1.1, Cesium 1.33 and prepare release

gberaudo:


gberaudo:
Thanks for the review @fredj.

PR #461, opened 2017-05-18
Building on Windows

tchandelle:
This PR updates `build/generate-info.js` to allow Windows user to build ol-cesium. I used [openlayers/tasks/generate-info.js](https://github.com/openlayers/openlayers/blob/46c610f35e0adb89f73bb49c3d472d126e28b63c/tasks/generate-info.js) as a reference


gberaudo:
Sorry I forgot your PR. Thanks for contributing it.

PR #478, opened 2017-07-28
Remove listeners created in raster synchronizer

oterral:
RasterSynchonizer class added events that was never removed when the layer was removed.


gberaudo:
Thanks @oterral. A quick grep at the code shows other places where the listeners are not tracked and unregistered. Feel free to have a look to it if you have time.

PR #480, opened 2017-07-31
Fix #479: listen specific events on list of layers instead of global …

### LINKED ISSUE #479: Layer's opacity is not updated.
In the case of you're using Cesium over the ol map , the visibility of the root group of the map is [set to false](https://github.com/openlayers/ol-cesium/blob/master/src/olcesium.js#L543) . This behavior, combined with the enabling of the auto render loop process,  prevents the layer's opacity  to be updated if the render loop is stopped.

If the root group has a visibility of false, [the change event is not propagated](https://github.com/openlayers/openlayers/blob/v4.2.0/src/ol/layer/group.js#L83)  and [catched by ol-cesium](https://github.com/openlayers/ol-cesium/blob/master/src/autorenderloop.js#L186)

You can reproduce it using the rastersync example in this zip file:
[rastersync.zip](https://github.com/openlayers/ol-cesium/files/1187048/rastersync.zip)

oterral:
…change event on the root layer group

Fix #479

### DIFF
--- a/src/autorenderloop.js
+++ b/src/autorenderloop.js
@@ -143,9 +143,16 @@ olcs.AutoRenderLoop.prototype.enable = function() {
     that.notifyRepaintRequired();
   };
 
-  // Listen for changes on the layer group
-  this.ol3d.getOlMap().getLayerGroup().on('change',
+  // Listen for changes on the main layer group
+  this.ol3d.getOlMap().getLayerGroup().on('change:layers',
       this._boundNotifyRepaintRequired);
+
+  this.ol3d.getOlMap().getLayers().forEach((layer) => {
+    // Listen for changes on layer group
+    layer.on('change', that._boundNotifyRepaintRequired);
+    // Listen for property changes on simple layer (opacity, visibility, ...)
+    layer.on('propertychange', that._boundNotifyRepaintRequired);
+  });
 };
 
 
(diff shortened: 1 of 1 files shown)


src/autorenderloop.js
```diff
@@ -143,9 +143,16 @@ olcs.AutoRenderLoop.prototype.enable = function() {
...
+  // Listen for changes on the main layer group
+  this.ol3d.getOlMap().getLayerGroup().on('change:layers',
       this._boundNotifyRepaintRequired);
+
+  this.ol3d.getOlMap().getLayers().forEach((layer) => {
+    // Listen for changes on layer group
+    layer.on('change', that._boundNotifyRepaintRequired);
+    // Listen for property changes on simple layer (opacity, visibility, ...)
+    layer.on('propertychange', that._boundNotifyRepaintRequired);
+  });
```
gberaudo: That will not work if a layer is added to the map after the autorender loop is enabled.
I am looking if there is another way.

gberaudo: @fredj, do you know if the optimization at https://github.com/openlayers/openlayers/blob/v4.2.0/src/ol/layer/group.js#L83 is necessary? Maybe it could be acceptable to relax this condition for the main group?

fredj: yes, this could probably be relaxed because we don't have this behavior with the ol.layer.Layer (changed() is called even if the layer is not visible)

fredj: @gberaudo see https://github.com/openlayers/openlayers/pull/899

oterral: @gberaudo, takes in account that this [change](https://github.com/openlayers/ol-cesium/commit/5537b310579378f81660d8d59e8d5935c40ea164) is also involved.

oterral: @gberaudo I can listen also add / remove events on the layers collection. What do you think ?

gberaudo: Yesterday I tried out removing the `if` in the OpenLayers code and it was responsive in OL3-Cesium. Looking at the OpenLayers code it looks more consistent to do the change in OpenLayers: https://github.com/openlayers/openlayers/pull/7073.

gberaudo:
Fixed by https://github.com/openlayers/openlayers/pull/7073 in OpenLayers.

PR #487, opened 2017-08-29
Add some externs

oterral:


Cesium.externs.js
```diff
@@ -2763,6 +2768,11 @@ Cesium.CustomDataSource.prototype.entities;
  */
 Cesium.EntityCollection = function() {}
 
+/**
+ * @param {string} id
```
gberaudo: According to Cesium code, it looks to me that it should be `String|Number`. I opened a tentative fix in Cesium, see https://github.com/AnalyticalGraphicsInc/cesium/pull/5791.

oterral: Yes makes sense.

gberaudo:
Let's see the outcome of the Cesium PR before merging.

Cesium.externs.js
gberaudo: It appears your previous attempt was the good one. Can you revert to it?

oterral: Done

gberaudo:
Thanks!

PR #488, opened 2017-08-30
Depend on OpenLayers and Cesium through npm

gberaudo:
Switch to @camptocamp/closure-util to allow usage of goog.modules.
Applications using closure will need to switch to @camptocamp/closure-util.
In the middle term the dependency on closure-util will be removed in favour of pure ES6 modules.

Switch to @camptocamp/cesium to allow writing examples using Cesium vector tiles or advanced optimizations.


gberaudo:
Thanks for the review.

PR #494, opened 2017-09-15
synchronize nested ol.layer.group Items

jbo023:
Hi, 
we are at the moment working on a project to integrate 3D in the 2D gis portal of Hamburg. For this we are using ol-cesium, but we had to add some functionality and now want to integrate the changes in the project. 
For better review purposes i will create several pull requests. 

Here we changed the code to also synchronize nested ol.layer.group Items. 

Link to the hamburg application: https://www.geoportal-hamburg.de/Geoportal/geo-online/
Link to the Hamburg application source code: https://bitbucket.org/lgv-g12/lgv

This work is paid by the Agency for Geoinformation and Surveying (LGV) Hamburg
http://www.hamburg.de/bsw/whoweare/

Jannes

PR #495, opened 2017-09-15
prevent click propagation from cesium to the underlying ol mapEvents

jbo023:
next PR for the city of hamburg project see https://github.com/openlayers/ol-cesium/pull/494 

if we use 3D with the same target, we don't want to trigger mouse events on the underlying 2D Map.
If someone wants to keep the old behavior we could maybe implement this behavior optional with a flag. 

Jannes

### DIFF
--- a/src/olcs/olcesium.js
+++ b/src/olcs/olcesium.js
@@ -3,6 +3,8 @@ goog.require('ol.geom.Point');
 
 goog.require('goog.asserts');
 goog.require('ol.proj');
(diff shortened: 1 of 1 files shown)


gberaudo:
(note: this PR content is part of https://github.com/openlayers/ol-cesium/pull/498)

gberaudo:
@jbo023, propagating the events is actually useful. See the new flag in https://github.com/openlayers/ol-cesium/pull/529.

Issue #493, opened 2017-09-15
is Layer Cluster  is doesn't work?


wYuanHpp:
when I use ol3 to build my project ,every thing is cool and worked.But when I try to  add olcesium to my project, I find my layer's Cluster  hasn't any feature , so my map doesn't display any feature.but this cluster's source has the right number features. I don't know why, help me thank you!

gberaudo:
Hi @wYuanHpp, at the moment there is no support for clustering in OL-Cesium.

PR #496, opened 2017-09-18
Do not use cluster source, but the underlying feature source

jbo023:
next PR for the city of hamburg project see #494

this fixes https://github.com/openlayers/ol-cesium/issues/493

we changed the handling of cluster sources. As a first iteration we use the real features to convert to cesium. 

As a second iteration someone could intergrate the cesium clustering support.

Jannes


gberaudo:
Thanks @jbo023.

PR #497, opened 2017-09-18
add nullcheck, computeplainstyle can return null

jbo023:
just fixes a small bug i saw when i did some of the other work. 

computeplainstyle can return null, so we have to check for null

Jannes


gberaudo:
Thanks

PR #498, opened 2017-09-20
Overlay Synchronization

jbo023:
next PR for the city of hamburg project see #494

we implemented an overviewsynchronizer


src/olcs/overlaysynchronizer.js
```diff
@@ -0,0 +1,169 @@
...
+   * @type {!ol.Map}
+   * @protected
+   */
+  this.map = map;
+
+  /**
+   * @type {ol.Collection.<ol.Overlay>}
+   * @private
+   */
+  this.overlays = this.map.getOverlays();
```
gberaudo: If it is private, it should have a trailing _.

src/olcs/overlaysynchronizer.js
```diff
@@ -0,0 +1,169 @@
...
+    ol.events.EventType.MOUSEWHEEL,
+    ol.events.EventType.WHEEL
+  ];
+  overlayEvents.forEach((event) => {
+    ol.events.listen(this.overlayContainerStopEvent_, event, ol.events.Event.stopPropagation);
+  });
+  this.scene.canvas.parentElement.appendChild(this.overlayContainerStopEvent_);
+
+  /**
+   * @type {!Object<?,olcs.Overlay>}
```
gberaudo: Shouldn't it be `{!Object<string,olcs.Overlay>}`?

src/olcs/overlay.js
```diff
@@ -0,0 +1,350 @@
...
+   */
+  this.scenePostRenderListenerRemover_ = null;
+
+  /**
+   * @type {!Cesium.Scene}
+   */
+  this.scene = options.scene;
+
+  /** @type {!olcs.OverlaySynchronizer} */
+  this.synchronizer = options.synchronizer;
```
gberaudo: These 2 properties should be private.

src/olcs/overlay.js
```diff
@@ -0,0 +1,350 @@
...
+   * @todo read from parent
+   */
+  this.insertFirst_ = true;
+
+  /**
+   * @private
+   * @type {boolean}
+   * @todo read from parent
+   */
+  this.stopEvent_ = true;
```
gberaudo: What are you going to do about the todos?

src/olcs/overlay.js
```diff
@@ -0,0 +1,350 @@
...
+   */
+  this.element_ = document.createElement('DIV');
+  this.element_.className = `ol-overlay-container ${ol.css.CLASS_SELECTABLE}`;
+  this.element_.style.position = 'absolute';
+
+  /**
+   * @private
+   * @type {!ol.Overlay}
+   */
+  this.parent_ = options['parent'];
```
gberaudo: The quotes should be removed, see comment above.

src/olcs/overlay.js
```diff
@@ -0,0 +1,350 @@
...
+    this.scenePostRenderListenerRemover_();
+  }
+  this.scenePostRenderListenerRemover_ = null;
+
+  const scene = this.getScene();
+  if (scene) {
+    this.scenePostRenderListenerRemover_ = scene.postRender.addEventListener(this.updatePixelPosition.bind(this));
+    this.updatePixelPosition();
+    const container = this.stopEvent_ ?
+      this.synchronizer.getOverlayContainerStopEvent() : this.synchronizer.getOverlayContainer(); // TODO respect stop-event flag in synchronizer
```
gberaudo: What about this todo?

src/olcs/overlay.js
```diff
@@ -0,0 +1,350 @@
...
+    this.setVisible(false);
+    return;
+  }
+  this.setVisible(true);
+
+  const pixelCartesian = scene.cartesianToCanvasCoordinates(cartesian);
+  const pixel = [pixelCartesian.x, pixelCartesian.y];
+  const mapSize = [scene.canvas.width, scene.canvas.height];
+  const that = /** @type {ol.Overlay} */ (this);
+  ol.Overlay.prototype.updateRenderedPosition.call(that, pixel, mapSize);
```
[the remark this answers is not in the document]
jbo023: Yeah we also thought about this, but for this to work we have to either change the Overlay property visibility to "protected" from "private", or we can disable the closure compiler "visibility" Errors like this in the olcesium.json. This is also why we didn't do some of the todos in the code.

"jscomp_off": [
      "lintChecks",
      "analyzerChecks",
      "visibility"
    ],

So, what is your preference, should i disable the compiler flag, or should i create a openlayers PR which changes all ol.Overlay "@private" to "@protected". Then i can refactor the olcs.overlay to directly inherit from ol.Overlay.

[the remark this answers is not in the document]
gberaudo: You should send a PR to OpenLayers to change the visibility to `@protected`. It is something we do when it appears there is a real use-case for it.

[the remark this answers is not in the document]
jbo023: Hmm, i just checked this, changing properties to @protected is unfortunately not enough because i need to access all the properties of the original Overlay in the new olcs.Overlay. To do this we have three options
- change propertie visibility ( insertFirst_, stopEvent_, ect.) to @public
- add getters properties which do not have them yet 
- add getOptions function which returns an options Object (my preferred  solution)

Which option do you prefer?

jbo023:
Hi, 

i did some refactoring of the overlay synchronization. I renamed it to synchronizedOverlay and now inherit from ol/overlay.js. I think i also fixed all of your remarks, if still applicable. 
Jannes

jbo023:
This only works after the referenced openlayers PR is merged. For tests the file /node_modules/openlayers/src/ol/overlay.js can be changed with the changes from the PR

Cesium.externs.js
```diff
@@ -3416,3 +3444,37 @@ Cesium.EntityView.prototype.update = function(currentTime, bs) {};
...
+
+
+/**
+ * @param {Array.<Cesium.Cartesian4>} planes
+ * @constructor
+ */
+Cesium.CullingVolume = function(planes) {};
+
+/**
+ * @param {Object} boundingVolume
```
gberaudo: The type should be `Cesium.BoundingSphere`.

gberaudo:
Thanks, great. Waiting for the OpenLayers PR to be merged.

Cesium.externs.js
jbo023: I took the api description from the cesium API: https://cesiumjs.org/Cesium/Build/Documentation/CullingVolume.html
So computeVisibility works with every object which implements intersectsPlane().

gberaudo: OK, then fine with it since it would be too cumbersome to introduce interfaces.

gberaudo:
@jbo023, could you please rebase and change package.json to depend on the updated OpenLayers version?
```
    "openlayers": "git+https://github.com/openlayers/openlayers#1a1d45fdd",
```

jbo023:
@gberaudo 

rebase done, and i changed the openlayers dependency.

gberaudo:
I will now merge, thanks @jbo023 for your contribution.

PR #502, opened 2017-10-17
Add 3D enable event observer

heavis:
Add on and un function to support event observable, and add event observer to listen enabled property change. we can listen enable change as below:
 this.olcesium.on('enable3d', () => {
      const enabled = self.olcesium.getEnabled();
    });


gberaudo:
Hi @heavis, what is your use-case?

fgravin:
Hi, 
I haven't seen this PR before, but seems https://github.com/openlayers/ol-cesium/pull/538 should respond to your need.

PR #504, opened 2017-10-23
extend Camera API

jbo023:
in the Hamburg project, we would like to use the olcs.camera.calcResolutionForDistance function so we don't have to implement the same functionality. 

This PR just exports calcResoltionForDistance and calcDistanceForResolution so functions can be used from a thirdparty library

### DIFF
--- a/src/olcs/camera.js
+++ b/src/olcs/camera.js
@@ -388,7 +388,7 @@ olcs.Camera.prototype.readFromView = function() {
   goog.asserts.assert(ll);
 
   const resolution = this.view_.getResolution();
-  this.distance_ = this.calcDistanceForResolution_(
+  this.distance_ = this.calcDistanceForResolution(
       resolution || 0, ol.math.toRadians(ll[1]));
 
   this.updateCamera_();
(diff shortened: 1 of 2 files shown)


src/olcs/camera.js
```diff
@@ -491,12 +491,14 @@ olcs.Camera.prototype.checkCameraChange = function(opt_dontSync) {
 
 
 /**
+ * calculate the distance between camera and centerpoint based on the resolution and latitude value
  * @param {number} resolution Number of map units per pixel.
  * @param {number} latitude Latitude in radians.
  * @return {number} The calculated distance.
- * @private
+ * @public
```
fredj: `@public`not needed (it's the default)

src/olcs/camera.js
```diff
@@ -531,13 +533,15 @@ olcs.Camera.prototype.calcDistanceForResolution_ = function(resolution, latitude
 
 
 /**
+ * calculate the resolution based on a distance(camera to position) and latitude value
  * @param {number} distance
  * @param {number} latitude
  * @return {number} The calculated resolution.
- * @private
+ * @public
```
fredj: `@public`not needed (it's the default)

jbo023:
@fredj i removed the public annotations and rebased the branch to master

Is there anything else you need before merging ?

fredj:
@gberaudo please merge if you're happy with this changes

gberaudo:
Thanks

PR #505, opened 2017-10-25
Manager and Lazy loader contrib

gberaudo:


gberaudo:
@fredj, I removed the debug flag and try/catch which were actually not necessary.

PR #510, opened 2017-10-30
Enhance and rename toggle3d

gberaudo:
- add animation;
- rename to lowercase.

### DIFF
--- a/src/olcs/contrib/manager.js
+++ b/src/olcs/contrib/manager.js
@@ -2,8 +2,10 @@ goog.provide('olcs.contrib.Manager');
 
 goog.require('olcs.contrib.LazyLoader');
 goog.require('olcs.OLCesium');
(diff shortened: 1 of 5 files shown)


src/olcs/contrib/manager.js
```diff
@@ -44,6 +45,12 @@ olcs.contrib.Manager = class {
      * @protected
      */
     this.ol3d;
+
+    /**
+     * @const {number} Tilt angle in radians
+     * @private
+     */
+    this.cesiumInitialTilt_ = olcs.core.toRadians(50);
```
fredj: Use `Cesium.Math.toRadians` (or `ol.math.toRadians`) instead

gberaudo: Thanks for the suggestion. I updated the PR.

PR #521, opened 2017-11-17
Constrain camera with bounding sphere

gberaudo:


gberaudo:
@fredj, I have updated the PR, can you have a new look?

PR #525, opened 2017-11-21
Small cosmetic changes

gberaudo:
More or less depends on https://github.com/AnalyticalGraphicsInc/cesium/pull/5993.

### DIFF
--- a/src/olcs/contrib/manager.js
+++ b/src/olcs/contrib/manager.js
@@ -173,6 +173,10 @@ olcs.contrib.Manager = class {
     // sense anyway, except for debugging
     scene.globe.depthTestAgainstTerrain = true;
 
+    // Use white instead of the black default colour for the globe when tiles are missing
+    scene.globe.baseColor = Cesium.Color.WHITE;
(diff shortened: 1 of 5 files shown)


src/olcs/contrib/manager.js
```diff
@@ -335,4 +335,14 @@ olcs.contrib.Manager = class {
       });
     });
   }
+
+  /**
+   * @protected
+   * @return {Cesium.Rectangle|undefined}
+   */
+  getCameraExtentRectangle() {
```
fredj: why not using a getter `get cameraExtentRectangle` ?

gberaudo: I tried on another class and got an error from GCC. (something like export is only allowed for global methods and properties).

fredj: ok

gberaudo:
@fredj, I rebase and enabled the camera workaround by default. Does it still looks good to you?

gberaudo:
Thanks for the reviews @fredj.

PR #528, opened 2017-11-24
Update externs

gberaudo:


gberaudo:
This is trivial. Merging.

PR #529, opened 2017-11-24
Stop OpenLayers events propagation with a flag

gberaudo:
The goal of OL-Cesium is to unify interacting with OpenLayers and Cesium.
Letting the events be propagated to OpenLayers allows to have a single listener in application code, designed around the OpenLayers API.

### DIFF
--- a/externs/olcsx.js
+++ b/externs/olcsx.js
@@ -12,6 +12,7 @@ var olcsx;
  *   target: (Element|string|undefined),
  *   createSynchronizers: (undefined|function(!ol.Map, !Cesium.Scene, !Cesium.DataSourceCollection): Array.<olcs.AbstractSynchronizer>),
  *   time: (undefined|function(): Cesium.JulianDate),
+ *   stopOpenLayersEventsPropagation: (boolean|undefined),
  *   sceneOptions: (Cesium.SceneOptions|undefined)
  * }}
  * @api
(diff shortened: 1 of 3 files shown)


externs/olcsx.js
```diff
@@ -12,6 +12,7 @@ var olcsx;
  *   target: (Element|string|undefined),
  *   createSynchronizers: (undefined|function(!ol.Map, !Cesium.Scene, !Cesium.DataSourceCollection): Array.<olcs.AbstractSynchronizer>),
  *   time: (undefined|function(): Cesium.JulianDate),
+ *   stopOpenLayersEventsPropagation: (boolean|undefined),
```
fredj: should be added bellow:
```
/**
 * @type {boolean|undefined}
 */
olcsx.OLCesiumOptions.prototype.stopOpenLayersEventsPropagation;

```

gberaudo: done, thanks

jbo023:
@gberaudo 
Yeah there are probably some cases where you want to have the events. 
But you have to be careful with the coordinates if you click on the 3D Map, because the clicked Point in 3D is different to the coordinates you get back from 2D.

gberaudo:
Yes you are right @jbo023. We have some code to fix these coordinates before handing them to the application code.
Ideally this fixing should be automatically drived by OL-Cesium, it is something I would like to improve in the future.

PR #535, opened 2017-12-08
Set the eslint as the root config file

oterral:
It's a proposal, When we build ol-cesium in a subfolder , It tries to apply the eslint config from the parent folder. 
Set the root porperty to true avoid this.


gberaudo:
Thanks.

PR #537, opened 2017-12-14
Make ol3d load promise always available

fgravin:
This PR tries to improve the Cesium lazy load promise mechanism to provide an always available fresh promise that is resolved when load is completed.

The promise is accessible with `getLoadPromise()` method. 

This way anyone can know when load is done without requesting the load.

### DIFF
--- a/src/olcs/contrib/manager.js
+++ b/src/olcs/contrib/manager.js
@@ -47,10 +47,24 @@ olcs.contrib.Manager = class {
     this.blockLimiter_ = false;
 
     /**
+     * @type {function(Promise.<olcs.OLCesium>)}
+     * @private
+     */
+    this.loadPromiseResolver;
+
+    /**
      * @type {Promise.<olcs.OLCesium>}
      * @private
      */
-    this.promise_;
+    this.loadPromise_ = new Promise((resolve) => {
+      this.loadPromiseResolver = resolve;
+    });
+
+    /**
+     * @type {boolean}
+     * @private
+     */
+    this.waitForLoad_ = false;
 
     /**
      * @type {olcs.OLCesium}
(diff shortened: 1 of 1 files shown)


src/olcs/contrib/manager.js
```diff
@@ -46,11 +46,25 @@ olcs.contrib.Manager = class {
...
-    this.promise_;
+    this.loadPromise_ = new Promise((resolve) => {
+      this.loadPromiseResolver = resolve;
+    });
+
+    /**
+     * @type {boolean}
+     * @private
+     */
+    this.waitForLoad_ = false;
```
gberaudo: I think `needLazyLoading = true` would sound better.

src/olcs/contrib/manager.js
```diff
@@ -46,11 +46,25 @@ olcs.contrib.Manager = class {
...
+     * @private
+     */
+    this.loadPromiseResolver;
+
     /**
      * @type {Promise.<olcs.OLCesium>}
      * @private
      */
-    this.promise_;
+    this.loadPromise_ = new Promise((resolve) => {
```
gberaudo: Here you only handle the resolve. You must also handle the reject case.

src/olcs/contrib/manager.js
```diff
@@ -104,11 +118,21 @@ olcs.contrib.Manager = class {
...
+      );
     }
-    return this.promise_;
+    return this.loadPromise_;
+  }
+
+  /**
+   * @return {Promise.<olcs.OLCesium>}
+   */
+  getLoadPromise() {
```
gberaudo: To keep the API smaller, I would rather add a parameter to the `load` method: `load(lazyLoad=true)`.
Then this new method would not be necessary.

src/olcs/contrib/manager.js
```diff
@@ -46,11 +46,25 @@ olcs.contrib.Manager = class {
      */
     this.blockLimiter_ = false;
 
+    /**
+     * @type {function(Promise.<olcs.OLCesium>)}
+     * @private
+     */
+    this.loadPromiseResolver;
```
fredj: private variable should ends with `_`

PR #538, opened 2017-12-18
Add 3d toggle & load events in manager

fgravin:
Add an event when 3D is toggled to the Cesium Manager.

```js
cesiumManager.on('change', (event) => {
  console.log('3D active: ', event.target.is3dEnabled());
});
```
Also add a `LOAD` event when cesium has been loaded.

### DIFF
--- a/src/olcs/contrib/manager.js
+++ b/src/olcs/contrib/manager.js
@@ -131,6 +131,7 @@ olcs.contrib.Manager = class extends ol.Observable {
     const scene = this.ol3d.getCesiumScene();
     this.configureForUsability(scene);
     this.configureForPerformance(scene);
+    this.dispatchEvent('load');
     return this.ol3d;
   }
 
--- a/src/olcs/contrib/manager.js
+++ b/src/olcs/contrib/manager.js
@@ -5,16 +5,19 @@ goog.require('olcs.OLCesium');
 goog.require('olcs.core');
 
 goog.require('ol.math');
+goog.require('ol.Observable');
 goog.require('goog.asserts');
 
 
-olcs.contrib.Manager = class {
+olcs.contrib.Manager = class extends ol.Observable {
   /**
    * @param {string} cesiumUrl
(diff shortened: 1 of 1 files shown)


src/olcs/contrib/manager.js
```diff
@@ -229,10 +234,12 @@ olcs.contrib.Manager = class {
         goog.asserts.assert(this.map);
         return olcs.core.resetToNorthZenith(this.map, scene).then(() => {
           ol3d.setEnabled(false);
+          this.dispatchEvent(new ol.events.Event(ol.events.EventType.CHANGE));
```
gberaudo: Better call the event `toggle`. Then it is fine to have it on the `manager` instead of `ol3d`.

fgravin: What do you mean ? `ol.events.EventType.CHANGE` to be `'toggle'` ? So i need to create a new type ? It's not in ol event types list.

Also i saw a function `changed` in `ol.Observable` that i would use..

src/olcs/contrib/manager.js
```diff
@@ -128,6 +133,7 @@ olcs.contrib.Manager = class {
     const scene = this.ol3d.getCesiumScene();
     this.configureForUsability(scene);
     this.configureForPerformance(scene);
+    this.dispatchEvent(ol.events.EventType.LOAD)
```
gberaudo: Please directly use strings. Usage of `ol.events.EventType` will soon go away.
In addition, I have the feeling the event will be dispatched too early (the manager will not yet be ready). It might be fine to put the dispatch in a `setTimeout` call.

fgravin: Ok so my last commit is useless too ? I just put 'toggle' as parameter ?

fgravin: > It might be fine to put the dispatch in a setTimeout call.

In my example it looks it's loaded on time, ol3d object is ready in my event callback, but i will secure it with a timeout if you think it's better.

PR #539, opened 2017-12-19
Add method to init 3d with a view without transition

fgravin:
New method in manger to set up a 3d view without transition.
Parameters are

- lon
- lat
- elevation
- heading
- pitch

### DIFF
--- a/src/olcs/contrib/manager.js
+++ b/src/olcs/contrib/manager.js
@@ -246,6 +246,41 @@ olcs.contrib.Manager = class extends ol.Observable {
 
 
   /**
+   * Enable ol3d with a view built from parameters.
+   *
+   * @export
+   * @param {number} lon
+   * @param {number} lat
+   * @param {number} elevation
+   * @param {number} headingDeg Heading value in degrees.
+   * @param {number} pitchDeg Pitch value in degrees.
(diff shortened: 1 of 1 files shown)


src/olcs/contrib/manager.js
```diff
@@ -245,6 +245,41 @@ olcs.contrib.Manager = class extends ol.Observable {
...
+   *
+   * @export
+   * @param {number} lon
+   * @param {number} lat
+   * @param {number} elevation
+   * @param {number} headingDeg
+   * @param {number} pitchDeg
+   * @returns {Promise<undefined>}
+   */
+  set3dWithView(lon, lat, elevation, headingDeg, pitchDeg) {
```
gberaudo: To make it in cleaner, let's pass destination and orientation as parameters here.

gberaudo: You are right Cesium is not loaded so disregard my suggestion.

fgravin:
Ok i have tested it further within a project and it works as expected.
The only concern i have is the type of the parameter, should we accept `undefined` or not ? I would say no.

@gberaudo please last check and merge if ok.
Thanks

gberaudo:
Thanks @fgravin, yes the parameter handling is good.

PR #541, opened 2018-01-03
Add ImageWMS sync support

fgravin:
Aims to support `ImageWMS` when it is possible. For that, an alternative `TileWMS` source is created from the `ImageWMS` one and is used to create the Cesium `ImageryLayer`.

Support is not possible when ImageWMS
- has no `url`
- use an `ImageLoadFunction`

No ref to original `ImageWMS` source seems to be need, as the event only watch for `TileWMS` state to be `'ready'`.

### DIFF
--- a/examples/imageWMS.html
+++ b/examples/imageWMS.html
@@ -1,0 +1,17 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE HTML>
+<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
+  <head>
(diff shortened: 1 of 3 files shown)


fgravin:
> To avoid taking down his server, could you please create a dedicated example using stacked view (like in main.html)?

@gberaudo i don't understand what you expect from me. H

fgravin:
> To avoid taking down his server, could you please create a dedicated example using stacked view (like in main.html)?

I've reverted the rasterSync example and created a new one as you menshionned. From now on, i use boundless geoserver url for the wms layer.

Please review again @gberaudo

PR #549, opened 2018-01-31
Fix source projection retrieval for WMS

fgravin:
Keep default behavior:

If no projection is defined in the layer via `layer.getSource().getProjection()` or `layer.getSource().get('olcs.projection')`, then use map projection for the layer (that should support it because it is the one used for openlayers).


fgravin:
@gberaudo please review, thanks

PR #557, opened 2018-02-14
Fixes #556 after OL 3.5.0 API changes

### LINKED ISSUE #556: Using the tileSize parameter on a ol.source.XYZ breaks the Cesium map
Hello,

I have been able to narrow down the issue I was having to a bug that can easily be reproduced.
Using the following code:
```
var nativeLayer = new ol.layer.Tile({
	source: new ol.source.XYZ({
		url: "https://a.tile.openstreetmap.org/{z}/{x}/{y}.png",
		maxZoom: 18,
		tileSize: [256, 256]
	})
});


const map = new ol.Map({
  layers: [
	nativeLayer
  ],
  target: 'map2d',
  view: new ol.View({
    center: [0, 0],
    zoom: 2
  })
});

const ol3d = new olcs.OLCesium({map});
ol3d.setEnabled(true);
```
Produce a map with only the top-most tile, and it doesn't load any tile as you zoom in:
image
Zooming in : 
image
If I remove the tileSize parameter, I get the map as it should be:
image
I guess the problem is here in the code:
https://github.com/openlayers/ol-cesium/blob/59c05de2cabaa1d952653c5cd2d0cb1163b9c4b8/src/olcs/core/olimageryprovider.js#L108

Turns out tg.getTileSize(0) returns [256, 256] and not 256 as the library seem to expect
Looks like it changed in version 3.5.0:
https://github.com/openlayers/openlayers/blob/8d1346c1b6707879cdbf88d780b473cc1a995baa/changelog/v3.5.0.md#support-for-non-square-tiles

Fabien

fnicollet:


gberaudo:
Thanks for reporting the issue and providing a patch.

In order to simplify the code, could you please use the `ol.size.toSize()` as suggested in https://github.com/openlayers/openlayers/blob/8d1346c1b6707879cdbf88d780b473cc1a995baa/changelog/v3.5.0.md#support-for-non-square-tiles?

In addition, a few lines below could you add dedicated code for handling `tileHeight`?

fnicollet:
Hello,

I also tested a WMTS provider and it looks like sometimes the value returned is a number, not ol.size object (maybe they didn't implemented everywhere?). So testing if the object is array-like sounds safe to me, what do you think?
I'll add the code for tileHeight to the PR

Fabien

gberaudo:
Thanks.

PR #558, opened 2018-02-14
Properly clear billboards on feature remove

fgravin:
For the meantime, during feature removal, only geometry of type `ol.geom.Point` were removing their corresponding potential billboard (only one cause it's a point).

But ol-cesium also manage multipoint and multi style, for example a `LineString` with a style with a `geometryFunction` returning a `MultiPoint`. In those cases, billboards were not cleared.

So the pool of billboard for a feature is now an array, and the array is cleared whatever the type of the geometry when the feature is removed.


fgravin:
@gberaudo please have a look.
Thanks

fgravin:
Please merge @gberaudo

Issue #435, opened 2017-01-26
What is possible and what is not?


strech345:
Hello, for me it is not really clear what functionalities i will get from ol3 and cesium.

**Does i have the full features of Cesium?** for example ...
- Loading different elevation models
- new 3D vector tiles (http://cesiumjs.org/2015/08/10/introducing-3d-tiles/)
- 3d models in gltf
- loading primitive
- Plugins (for example cesium-navigaton https://github.com/alberto-acevedo/cesium-navigation)
- Interactions

**Does it have the full features of OL3 ?** for example ...
- Heatmaps
- Clustering of Marker
- loading Vector Tiles
- loading KML, Json,...
- Interactions

Would be great to get a comsumption about what is possible and what's not.

jmgomezpoveda:
A lot of things work. Now, I don't know about most of these, but I do know about a few of them:

* Interactions do not work; you have to use Cesium calls for that
* KML works, and clustering of markers, though there are some glitches (https://github.com/openlayers/ol3-cesium/issues/345, https://github.com/openlayers/ol3-cesium/issues/346)

gberaudo:
You create and pass the OL map so you have full control on it.
The Cesium globe is created by OL3-Cesium but you get notably the reference to the Cesium.Scene, you can then use the Cesium API.

For missing features it is possible to:
- improve OL3-Cesium by contributing patches;
- use hooks to implement own logics (eg synchronize a raster layer in OL3 to a vector layer in Cesium / gltf ...)

strech345:
@gberaudo 
for me it is important to have the reference on cesium viewer also and to control it like i want. It would be nice to create boath instances (ol3 and cesium) and to use this framework to connect boath with each other. But i think this will not be the case.

@jmgomezpoveda
So. its possible with cesium feature request to request also ol3 features?

jmgomezpoveda:
@strech345 I am not sure I understand your question. With ol3-cesium you use the same API as ol3, and ol3-cesium internally translates the API requests into the Cesium API. You create the map once and you are able to switch between the ol3 and Cesium views at will.

In most cases you only need to use the ol3 API for this. In some cases, though, you may need to interact directly with the Cesium object exposed by ol3-cesium.

Semyon62:
What Cesium calls I need to use to ol translate and select interactions?

PR #559, opened 2018-02-14
Add basic model conversion support

### LINKED ISSUE #211: 3D models synced with open layers?
I saw #163 . What I'm trying to do:
1. Make a feature, like a point, in Openlayers.
2. Set a key value pair with ol.Feature.set('model', '/models/airplane.gltf'). Maybe with some basic options like minimumPixelSize and scale.
3. When switching from Openlayers to Cesium, it replaces the point with that model.

Is something like having a 3D model synced (i.e. position) with a point in Openlayers possible? If so, can I please get an example? Thank you.

gberaudo:
Closes #211.
Related to #435.

image


examples/vectors.js
```diff
@@ -44,6 +46,14 @@ const cervinFeature = new ol.Feature({
 cervinFeature.getGeometry().set('altitudeMode', 'clampToGround');
 
 
+const modelFeatures = [-1, -1 / 2, 0, 1 / 2, 1, 3 / 2].map(
+    factor => new ol.Feature({
+      geometry: new ol.geom.Point([852641, 5776749, 4500]),
+      'rotation': factor * Math.PI
```
fredj: remove quotes?

gberaudo: I like to be ready for GCC compilation of the examples :P

gberaudo:
@fredj, as discussed yesterday, I changed the host of the `olcs_model` object to be an ol.Object.
Does it look good to you?

gberaudo:
Thanks for the rewiev @fredj.

PR #562, opened 2018-03-02
Add client side raster reprojection support

### LINKED ISSUE #37: Additional map projections (SRS) for raster formats
Cesium supports several raster formats, notably:
- WMTS;
- WMS.

In ol3cesium, we are currently restricted to a custom ol3 tile image provider.
This imagery provider gets the tiles from an ol.source.TileImage in either 4326 or 
3857 projection.

Other or different projections in ol3 and Cesium are not possible.
WMS or WMTS are not handled.

gberaudo:
Fixes #37.

### DIFF
--- a/src/olcs/core/olimageryprovider.js
+++ b/src/olcs/core/olimageryprovider.js
@@ -183,7 +183,7 @@ olcs.core.OLImageryProvider.prototype.handleSourceChanged_ = function() {
       this.tilingScheme_ = new Cesium.GeographicTilingScheme();
     } else if (this.projection_ == ol.proj.get('EPSG:3857')) {
       this.tilingScheme_ = new Cesium.WebMercatorTilingScheme();
-    } else if (ol.ENABLE_RASTER_REPROJECTION) {
+    } else if (ol.ENABLE_RASTER_REPROJECTION && olcs.core.OLImageryProvider.ENABLE_RASTER_REPROJECTION) {
       this.tilingScheme_ = new Cesium.GeographicTilingScheme();
       this.projection_ = ol.proj.get('EPSG:4326'); // reproject
     } else {
@@ -279,3 +279,9 @@ olcs.core.OLImageryProvider.prototype.requestImage = function(x, y, level) {
     return promise;
   }
 };
+
+/**
+ * @type {boolean}
+ * @export
+ */
+olcs.core.OLImageryProvider.ENABLE_RASTER_REPROJECTION = false;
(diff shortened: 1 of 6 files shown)


gberaudo:
@klokan, your feedbacks would be welcomed.

@ahocevar, we discussed a long time ago about this one. Thanks for your guidances back then at FOSS4G Seoul. If you find some time, a second look from you would be very appreciated.

src/olcs/core/olimageryprovider.js
```diff
@@ -247,23 +252,46 @@ olcs.core.OLImageryProvider.prototype.getTileCredits = function(x, y, level) {
...
-    return url ? Cesium.ImageryProvider.loadImage(this, url) : this.emptyCanvas_;
+  // Perform mapping of Cesium tile coordinates to ol3 tile coordinates:
+  // 1) Cesium zoom level 0 is OpenLayers zoom level 1 for EPSG:4326
+  const z_ = this.tilingScheme_ instanceof Cesium.GeographicTilingScheme ? level + 1 : level;
+  // 2) OpenLayers tile coordinates increase from bottom to top
+  const y_ = -y - 1;
+
+  const tilegrid = this.source_.getTileGridForProjection(this.projection_);
+  if (z_ < tilegrid.getMinZoom() || z_ > tilegrid.getMaxZoom()) {
+    return this.emptyCanvas_; // no data
```
wallw-bits: There are several places in this function which do not follow the Cesium API reference and return something other than `Promise<Image|Canvas>|undefined`. Is that intentional based on undocumented Cesium workings?

gberaudo: Hi @wallw-bits, you are right it works but it is incorrect. Can you come up with a PR wrapping the returned value like this: `Promise.resolve(this.emptyCanvas_)`?

src/olcs/core/olimageryprovider.js
```diff
@@ -247,23 +252,46 @@ olcs.core.OLImageryProvider.prototype.getTileCredits = function(x, y, level) {
...
+  // not yet loaded!
+  // const image = tile.getImage();
+  // if (!image || !image.src) {
+  //   return Promise.resolve(this.emptyCanvas_); // no data
+  // }
+
+
+  const state = tile.getState();
+  if (state === ol.TileState.LOADED || state === ol.TileState.EMPTY) {
+    return tile.getImage() || undefined;
```
wallw-bits: This return also needs a promise wrapper if `tile.getImage()` is not undefined

gberaudo: Thanks, done.

gberaudo:
I am hiding the feature behind a flag so that it can be merged. @fredj, could you please have a look to it?

gberaudo:
This branch have issues. I am restoring it and "unmerging" the PR.

PR #565, opened 2018-03-05
Fix double js suffix in olcs and other fixes

gberaudo:
With these changes the ES6 olcs package is in much better shape.
Almost a beta now!


gberaudo:
@fredj or @sbrunner, can you re-review?

build/package.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+rm -rf .build/package
+cp -r package .build
+(cd ./src && cp -r olcs/* ../.build/package)
```
sbrunner: => `cp -r src/olcs/* .build/package`

gberaudo: Indeed, will do in a future PR.

Issue #575, opened 2018-03-26
Display correct credit information


SangeethaS94:
In the latest version of ol-cesium, the credit information for the layers are not displayed correctly. The text, imageUrl and link parameters have been removed from Cesium.Credit in Cesium version 1.41 onwards. In the latest version of ol-cesium, three arguments are being passed when Credit object is created in function olcs.core.OLImageryProvider.createCreditForSource. In order to display the correct credit information, it must instead be one object that contains these three parameters.

gberaudo:
Hi @SangeethaS94, I fixed the parameter passed to Cesium and opened #576 with the change. However there looks to still be some issues.

I will not have time in the next 3 weeks to look at this. If you investigate this, you are welcome to provide a pull request.

gberaudo:
See also https://github.com/AnalyticalGraphicsInc/cesium/pull/6331.

SangeethaS94:
Hi @gberaudo!
Thanks for looking into it. 

In my case, I changed

`return (imageUrl || text.length > 0) ? new Cesium.Credit(text, imageUrl, link) : null;`

to 

```
var options = new Object();
options.text = text;
options.imageUrl = imageUrl;
options.link = link;
return (imageUrl || text.length > 0) ? new Cesium.Credit(options) : null;
```

And everything seems to work well for me. I am not getting the issues you have mentioned in [#576](https://github.com/openlayers/ol-cesium/pull/576)

gberaudo:
The issue I mentioned actually appears when the 3D globe is loaded first (not OpenLayers). This issue is unrelated to the OL and Cesium changes.

PR #576, opened 2018-03-27
Fix credit creation

gberaudo:
Related to #575.

I tested the examples, notably main is quite strange:
- I do not see the Cesium credits;
- instead there is an empty grey box (from OpenLayers);
- when toggling off then on the 3d, then credits show up (in an OpenLayers div...).

In top of that, Cesium is changing how it handles credits in next version (due next week).


gberaudo:
Handledi in #578.

PR #581, opened 2018-05-02
Client side raster reprojection

gberaudo:
Restore #562.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # Changelog
 
+# v 1.37
+
+* Changes
+  * Add client side raster reprojection.
+
 # v 1.36
 
 * Changes
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 # v 1.37
 
 * Changes
-  * Add client side raster reprojection.
+  * Add client side raster reprojection, see the example to see how to enable it.
 
 # v 1.36 - 2018-04-18
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@
 * Changes
   * Add client side raster reprojection.
 
-# v 1.36
+# v 1.36 - 2018-04-18
 
 * Changes
   * Rework the autorenderloop using Cesium RenderMode.
--- a/Cesium.externs.js
+++ b/Cesium.externs.js
@@ -2133,7 +2133,7 @@ Cesium.ImageryProvider.prototype.getTileCredits = function(x, y, level) {};
  * @param {number} x The tile X coordinate.
  * @param {number} y The tile Y coordinate.
  * @param {number} level The tile level.
- * @return {Object|undefined}
+ * @return {Promise<?>|undefined}
  */
 Cesium.ImageryProvider.prototype.requestImage = function(x, y, level) {};
 
--- a/examples/reprojection.html
+++ b/examples/reprojection.html
@@ -1,0 +1,17 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE HTML>
+<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
+  <head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+    <meta name="robots" content="index, all" />
+    <title>ol3cesium example</title>
+    <link rel="stylesheet" href="../node_modules/openlayers/css/ol.css" type="text/css">
+  </head>
+  <body>
+    <div id="map" style="width:600px;height:400px;"></div>
+    <input type="button" value="Enable/disable" onclick="javascript:ol3d.setEnabled(!ol3d.getEnabled())" />
+    <script src="inject_ol_cesium.js"></script>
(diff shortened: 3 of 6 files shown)


gberaudo:
There is an issue with this branch and the ImageWMS example.

wallw-bits:
We were interested something similar, not for raster reprojection itself but just getting Cesium to use the OL tile load stack. That way, implementing a custom tile layer (such as GeoPackage) only needs to be done in OpenLayers and Cesium will run through the same stack.

Our PR for the work on that is [here](https://github.com/ngageoint/opensphere/pull/257), with the relevant `ImageryProvider` changes [here]. Just a couple of notes compared with what you have here:

- Do not assume that the minimum zoom level of the tile grid is 0 for the `tileWidth` and `tileHeight` getters. Use `tg.getMinZoom()` instead.
- I tried the logic here for tiles in the `ol.TileState.EMPTY` state (in an overlay tile layer) which caused completely transparent tiles (base map layer should have shown). We went with making that state its own case.

Additionally, I went ahead and made a [TilingScheme](https://github.com/wallw-bits/opensphere/blob/cesium-use-ol-tiles/src/plugin/cesium/tilegridtilingscheme.js) implementation that uses the original OpenLayers `TileGrid` instead directly. This allows for more (but not all) of the non-typical cases handled by the `TileGrid` class. This has the added advantage of not having to adjust the zoom level by 1 for the `GeographicTilingScheme` versus the `WebMercatorTilingScheme`.

Note that the custom `TilingScheme` requires Cesium >1.46.

gberaudo:
Thanks @wallw-bits for the infos. It is pretty cool!

samuel-girard:
Hi,  

I am trying to rebase this branch onto ol-cesium v2 (currently working on the branch https://github.com/samuel-girard/ol-cesium/tree/olcs2-raster-reprojection).  
Everything works well, except for 2 examples:
- `imageWMS`: half of the globe is fine, the other half is black, not even loading OSM
- `customProj`: the globe is all black

I get in fact the same results than I faced last time (screenshots here: https://github.com/openlayers/ol-cesium/issues/37#issuecomment-370206118)  

It turns out these are the examples using an image source (`ol/source/Image.js`), other examples using tiled sources (`ol/source/Tile*`), I don't know if it is related though.

wallw-bits:
I completed this work for OpenSphere in these two files:
* https://github.com/ngageoint/opensphere/blob/master/src/plugin/cesium/imageryprovider.js
* https://github.com/wallw-bits/opensphere/blob/cesium-use-ol-tiles/src/plugin/cesium/tilegridtilingscheme.js

Those classes require Cesium >= 1.46.

You can also check out my [last comment above](https://github.com/openlayers/ol-cesium/pull/581#issuecomment-397760078) which has the details.

samuel-girard:
Hi @wallw-bits  
Thanks for your advice.  
I started to integrate the changes you did in opensphere for the reprojection (branch available here: [https://github.com/samuel-girard/ol-cesium/tree/raster-reprojection](https://github.com/samuel-girard/ol-cesium/tree/raster-reprojection)), but I am facing the exact same problem than with the previous work in ol-cesium: everything works well except for the examples `customProj` (globe is totally black) and `imageWMS` (only half of the globe has OSM map).

So far, I have no idea why I get these problems, but I will keep investigating.

wallw-bits:
You still have assumptions of a zero-level zoom in the `tileWidth` and `tileHeight` getters in `OLImageryProvider`. Also, [this line](https://github.com/samuel-girard/ol-cesium/blob/raster-reprojection/src/olcs/core/OLImageryProvider.js#L114) should not use an `instanceof <SomeTilingScheme>` to check the projection. Use `this.tilingScheme_.projection instanceof CesiumGeographicProjection` instead.

However, neither of those are really your problem in those examples. The changes to OpenSphere were to support tile layers. Both of the examples you are referencing use image layers, which do not use `TileGridTilingScheme` or `OLImageryProvider`. I think your problem is [here](https://github.com/samuel-girard/ol-cesium/blob/raster-reprojection/src/olcs/core.js#L416). That blurb is assuming that the source projection is `EPSG:4326`. You need to convert `source.getImageExtent()` from the source projection to `EPSG:4326` first. If that extent is in `EPSG:3857`, then the image is already being reprojected. If it is in `EPSG:21781`, then you will need to have OpenLayers reproject it first before showing it.

samuel-girard:
Thanks @wallw-bits for your feedback.

> I think your problem is [here](https://github.com/samuel-girard/ol-cesium/blob/raster-reprojection/src/olcs/core.js#L416). That blurb is assuming that the source projection is `EPSG:4326`. You need to convert `source.getImageExtent()` from the source projection to `EPSG:4326` first. If that extent is in `EPSG:3857`, then the image is already being reprojected. If it is in `EPSG:21781`, then you will need to have OpenLayers reproject it first before showing it.

The part you are referring to is about `Static` image sources (see the condition [here](https://github.com/samuel-girard/ol-cesium/blob/raster-reprojection/src/olcs/core.js#L406)). It wouldn't be the cause of the black globe issue (but I agree that it is another issue).

The problem I am facing is about `ImageWMS` sources (as you said, both failing examples use image layers), which are converted to `TileWMS` source [here](https://github.com/samuel-girard/ol-cesium/blob/0c20c26f9ce1d5651ebdde5bfe38ae1e769f46cf/src/olcs/core.js#L372).
Even if I add these layers as `TileWMS` source using a tiled layer, the problem is still here.

PR #600, opened 2018-06-18
Always rely on frameState to create attributions

dodecaphonic:
Some TileSources (TileUTFGrid, TileJSON and BingMaps, to date) define
their `setAttributions` function with a `frameState` argument,
provided by PluggableMap's POSTRENDER event in OpenLayers. With
ol-cesium, OLImageProvider calls `setAttributions` on instantiation,
without any arguments -- which, of course, results in an exception.

In this commit, we've implemented getTileCredits in order to
dynamically generate credits based on what's in the viewport.

An example using Bing (with the same credentials in OpenLayer's
repository) has been added to show the change in action.

### DIFF
--- a/examples/bing.html
+++ b/examples/bing.html
@@ -1,0 +1,16 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE HTML>
+<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
+  <head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+    <meta name="robots" content="index, all" />
+    <title>Bing example</title>
+    <link rel="stylesheet" href="../node_modules/ol/ol.css" type="text/css">
+  </head>
+  <body>
+    <div id="map" style="width:600px;height:400px;"></div>
+    <input id="enable" type="button" value="Enable/disable" />
+
+    <script src="inject_ol_cesium.js"></script>
+  </body>
+</html>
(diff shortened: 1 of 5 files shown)


dodecaphonic:
Hi @gberaudo,

Great. I didn't have all that context when I made the change. You are absolutely right.

I'll look into what you've suggested and adapt my patch accordingly.

gberaudo:
Hi @dodecaphonic, what is the status of your PR? Are you still working on it?

There is interest in having a resolution for this, see https://github.com/openlayers/ol-cesium/issues/607#issuecomment-406190586.

gberaudo:
After a quick look to the code, here is my suggestion:
In src/olcs/core/OLImageryProvider.js
- remove the createCreditForSource function and remove the this.credit_ property;
- implement getTileCredits(x, y, level) with something like:
```js
const zoom = this.tilingScheme_ instanceof Cesium.GeographicTilingScheme ? level + 1 : level;
const attributions = this.source_.getAttributions();
const olExtent = this.map_.getView().calculateExtent(this.map_.getSize());
const framestate = {
  viewState: {zoom},
  extent: olExtent
};
const credits = attributions.map(olAttribution => createCreditFromAttribution(attribution(framestate));
return credits;
```

dodecaphonic:
Hi, @gberaudo. I'm getting back to it in a couple of days and will look into your suggestions. I need to fix this, since my project still depends on olcs.

dodecaphonic:
@gberaudo I've made the changes you've suggested, and it seems to be working correctly.

gberaudo:
Thanks @dodecaphonic for your contribution.

dodecaphonic:
Thank you for your guidance, @gberaudo.

Issue #593, opened 2018-05-29
Stacked ground corridors for polygon outline


ediebold:
The stacked ground corridors work great for polylines, but aren't implemented for polygon outlines. It'd be great if we could get this implemented in the same way.

gberaudo:
Yes I guess it is possible to take all rings of a grounded polygon and create stacked ground corridors from them. PR welcome.

ediebold:
Thanks. I'd had a quick look, but was scared off by the different structure of the polygon conversion code. I really need it though, so once I can get my issues in #594 fixed I'll give it a go.

ediebold:
I was about to try and do this, but it looks like the next version of Cesium is going to include support for polylines on terrain: https://github.com/AnalyticalGraphicsInc/cesium/pull/6615

With that in mind, it probably makes more sense to wait for the next Cesium release and simply replace the stacked corridor code with support for the native polylines on terrain. After doing that, it shouldn't be too hard to use the same system for the polygon outlines.

So once the next Cesium release is out, I'll get to work on a PR for this.

PR #603, opened 2018-07-04
Added support for GroundPolylineGeometry in both linestrings and polygon outlines

ediebold:
As discussed in https://github.com/openlayers/ol-cesium/issues/593, I've made changes to FeatureConverter so that it uses Cesium's new GroundPolylineGeometry and GroundPolylinePrimitive. While we wait for Cesium to add this kind of support to polygon outlines on their own end (https://github.com/AnalyticalGraphicsInc/cesium/issues/6694), I've also added some code to the polygon converter, so that when ground clamping is on, it renders a polyline for the outline of each ring. 

All of this can be viewed in the new ground vectors example I made, which is mostly a trimmed version of the main vectors example.

Some other notes:

- Testing this requires Cesium 1.47. camptocamp's fork is only up to 1.45 at the moment, so I changed the devdependency locally when I was working on this. (Side note: what are the differences in this fork? I couldn't find any details about what's different).
- I've left the "createStackedGroundCorridors" code untouched, but I don't think it's being used anywhere in this branch now. Should I delete the function?


src/olcs/FeatureConverter.js
```diff
@@ -213,7 +213,7 @@ exports.prototype.wrapFillAndOutlineGeometries = function(layer, feature, olGeom
     primitives.add(p1);
   }
 
-  if (olStyle.getStroke()) {
+  if (olStyle.getStroke() && outlineGeometry) {
```
ediebold: I added this extra check because the code that creates the polylines for polygon outlines never sets outlineGeometry. This is another bit of code that can be removed when Cesium adds support for PolygonOutlines on terrain natively.

src/olcs/FeatureConverter.js
```diff
@@ -495,16 +504,55 @@ exports.prototype.olPolygonGeometryToCesium = function(layer, feature, olGeometr
...
+        polygonHierarchy: hierarchy,
+        perPositionHeight: true
+      });
+    }
   }
 
   const primitives = this.wrapFillAndOutlineGeometries(
       layer, feature, olGeometry, fillGeometry, outlineGeometry, olStyle);
 
+  if (typeof outlinePrimitives !== 'undefined') {
```
gberaudo: You declare the variable at the top of the function so you can simply do
```
if (outlinePrimitives) {...
```

ediebold:
Done!

I left the toggle ground clamping button in though, since I think it's actually useful to show the difference.

jbo023:
Hi, 

you should keep the old corridor code as a fallback option in case of the ie11. The Internet Explorer 11 unfortunatly does not support the cesium implementation, you can check with 
Cesium.GroundPolylinePrimitive.isSupported(this.scene)

ediebold:
Thanks @jbo023, I didn't know about this limitation. Shouldn't be too much work to put the stacked corridor code as a fallback - but I won't be able to test it, since I only have a Mac. If I put something together, would you be able to test the fallback works as expected?

jbo023:
yeah I can check, but I didn't setup the new dev environment with for ol-cesium2.0/ol5.0 yet so it can take some time

src/olcs/FeatureConverter.js
ediebold: changed

ediebold:
Okay, I've made it so stacked corridors are used when ground polylines aren't supported. I tested it locally by just adding "&& false" to the end of the relevant if statements, and it seems to work (plus, polygon outlines are now supported by the stacked corridors!)

bkuster:
Hi there,
I've been working on a similar issue and would like to raise two concerns:

1.  Is there a specific reasoning for creating a primitive per polygon outline  [here](https://github.com/openlayers/ol-cesium/pull/603/files#diff-bd5f86daf411ebcaa6cd437b67649139R519) (eg. for each hole) instead of just creating one primitive with an Array of geometry instances?
2. I've had trouble getting the picking behavior to work, since the `GroundPolylinePrimitive` is in fact _not_ a primitive, but a wrapper. I had to hack something like this into my code:
```javascript
outlinePrimitive.readyPromise.then(() => {
  this.setReferenceForPicking(layer, feature, outlinePrimitive._primitive);
});
```

ediebold:
Thanks for the feedback @bkuster.

1. I actually didn't know this was something that could be done. You're right that it makes way more sense. I'm going to update it now.
2. I hadn't thought about picking. Thanks for already having found that trick. I'll add that now as well and give it a quick test.

src/olcs/FeatureConverter.js
```diff
@@ -340,10 +340,14 @@ exports.prototype.olCircleGeometryToCesium = function(layer, feature, olGeometry
  * @param {!Array<Cesium.Cartesian3>} positions The vertices of the line.
  * @return {!Cesium.GroundPrimitive} primitive
  */
-exports.prototype.createStackedGroundCorridors = function(layer, feature, width, color, positions) {
-  let previousDistance = 0;
+exports.prototype.createStackedGroundCorridors = function(width, color, positions) {
```
gberaudo: You should keep the `layer and `feature` parameters.
They allow projects overriding the `createStackedGroundCorridors` method to have different behaviour depending on the actual layer/feature being converted.

src/olcs/FeatureConverter.js
```diff
@@ -340,10 +340,14 @@ exports.prototype.olCircleGeometryToCesium = function(layer, feature, olGeometry
...
+  // Convert positions to an Array if it isn't
+  if (positions[0].constructor.name !== 'Array') {
```
gberaudo: This looks inconsistent with the declared type of the `positions` parameter (which is an array of Cartesian3.
I believe you want to change the type to `!Array<Cesium.Cartesian3>|Array<Array<Cesium.Cartesian3>>`

Please also use the standard `Array.isArray()` function.

src/olcs/FeatureConverter.js
```diff
@@ -352,17 +356,19 @@ exports.prototype.createStackedGroundCorridors = function(layer, feature, width,
...
     };
-    geometryInstances.push(new Cesium.GeometryInstance({
-      geometry: new Cesium.CorridorGeometry(geometryOptions),
-      attributes: {
-        color: Cesium.ColorGeometryInstanceAttribute.fromColor(color),
-        distanceDisplayCondition: new Cesium.DistanceDisplayConditionGeometryInstanceAttribute(previousDistance, distance - 1)
-      }
-    }));
+    for (const linePositions of positions) {
+      geometryOptions.positions = linePositions;
```
gberaudo: It looks strange to change the `geometryOptions` object like this. You should move here the creation of the object so that a new object is created at each iteration of the loop.

src/olcs/FeatureConverter.js
```diff
@@ -404,15 +410,29 @@ exports.prototype.olLineStringGeometryToCesium = function(layer, feature, olGeom
...
+    let geometry;
+    const primitiveOptions = {
       // always update Cesium externs before adding a property
-      geometryInstances: new Cesium.GeometryInstance({
-        geometry: new Cesium.PolylineGeometry(geometryOptions)
-      }),
       appearance
-    });
+    };
+    if (heightReference == Cesium.HeightReference.CLAMP_TO_GROUND) {
```
gberaudo: It is less magics to use `===`. Would you be able to fix it here and at other places in this file?

src/olcs/FeatureConverter.js
```diff
@@ -404,15 +410,29 @@ exports.prototype.olLineStringGeometryToCesium = function(layer, feature, olGeom
...
+      geometry = new Cesium.GroundPolylineGeometry(geometryOptions);
```
gberaudo: Can you remove the `let geometry` above and change to `const geometry = ...` here and in the else below?

src/olcs/FeatureConverter.js
```diff
@@ -495,16 +517,63 @@ exports.prototype.olPolygonGeometryToCesium = function(layer, feature, olGeometr
...
+          outlineGeometry = this.createStackedGroundCorridors(width, color, positions);
+        } else {
+          const appearance = new Cesium.PolylineMaterialAppearance({
+            // always update Cesium externs before adding a property
+            material: this.olStyleToCesium(feature, olStyle, true)
+          });
+          const primitiveOptions = {
+            // always update Cesium externs before adding a property
+            appearance
+          };
```
gberaudo: Now that only one primitive is created you should move the object creation after the loop.

src/olcs/FeatureConverter.js
ediebold: That makes sense.

src/olcs/FeatureConverter.js
ediebold: Interesting- I was specifically trying to avoid recreating the rest of the object each time. I assumed there would be a performance improvement that way.  I'll change it now for consistency though.

ediebold:
@gberaudo - I believe I've made all the changes you requested

src/olcs/FeatureConverter.js
gberaudo: The problem is that your previous code relied on an implementation detail in Cesium. If Cesium were to modify its behaviour our code would break.

ediebold: Isn’t that the case either way? Although I guess it’s easier to find the spot to change things if the object is made in one spot.

gberaudo:
@ediebold, there seems to be an issue with the fallback when ground polyline primitives are not supported. When I force that code to be used and I don't see anymore the lines on the ground (I only see the filled polygons).
Could you please check?

gberaudo:
Thanks again @ediebold, @jbo023 and @bkuster. It is pretty cool to have this long awaited feature merged in. It will be part of the soon to be released v2.1.

PR #604, opened 2018-07-05
Port to Cesium 1.47

gberaudo:


gberaudo:
I tested locally. I am merging it so that we can go forward with #603.

PR #605, opened 2018-07-12
added circle outline GroundPolyline

bkuster:
Taking the ideas of @ediebold used in #603, I added support for clamped  `ol.geom.Circle` outline geometries.

### DIFF
--- a/src/olcs/FeatureConverter.js
+++ b/src/olcs/FeatureConverter.js
@@ -5,7 +5,7 @@ import olGeomGeometry from 'ol/geom/Geometry.js';
 import olStyleIcon from 'ol/style/Icon.js';
 import olSourceVector from 'ol/source/Vector.js';
 import olSourceCluster from 'ol/source/Cluster.js';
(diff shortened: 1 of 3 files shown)


src/olcs/FeatureConverter.js
```diff
@@ -5,6 +5,7 @@ import olGeomGeometry from 'ol/geom/Geometry.js';
 import olStyleIcon from 'ol/style/Icon.js';
 import olSourceVector from 'ol/source/Vector.js';
 import olSourceCluster from 'ol/source/Cluster.js';
+import {circular} from 'ol/geom/Polygon.js';
```
gberaudo: Please store it as `olCreateCircularPolygon`, so that it is clear at first glance what the function does and from which library the function comes.

bkuster: sure thing, changed in 7b01d9752dc7ba7edf00832dd6ae39294945363f

PR #609, opened 2018-07-16
Switch to ES6 classes

gberaudo:

PR #611, opened 2018-07-16
Fix SynchronizedOverlay with ES6 classes

gberaudo:
#609 broke synchronized overlays due to:
- a bad commit;
- the call to super that moved up (race condition).


gberaudo:
Merging this for the release. Tested with the overlay example.

PR #627, opened 2018-08-07
Fix missing source issue

gberaudo:
Closes: #626.


gberaudo:
Thanks @samuel-girard, will merge when Travis is done.

PR #628, opened 2018-08-08
Export module in package.json

samuel-girard:
Hello,
For any application using Webpack to build, we should point the `"module"` property (package.json) to the library source instead of its built version.

It also solves a problem that started to occur in ol-cesium 2.2: I got an error "_ol is not defined_" in the browser (I did not investigate further, but seems to be related to the "oldfashioned" library change from the 2.2 version).


samuel-girard:
That's great. Thank you @gberaudo

Issue #636, opened 2018-08-28
 Can't resolve 'goog/asserts.js'


felix9ia:
when I 
```
import OLCesium from 'ol-cesium';
```
but i got this
```
./node_modules/ol-cesium/src/olcs/OLCesium.js
Module not found: Can't resolve 'goog/asserts.js' in 'D:\sourcecode\catalog2.0-react\node_modules\ol-cesium\src\olcs'
```

version:  2.2.2
win10
node: 8.4

gberaudo:
Currently you need a webpack alias to hande the `goog` namespace.

felix9ia:
Thanks!Understand, but I still use the reference method of <script> on my React project. It seems that the support for webpack is not very good now. When I run `npm run build`, there will be an error. 
```
0.570d21b1.async.js from UglifyJs
Unexpected token: punc ()) [0.570d21b1.async.js:13245,19]
```
It seems that uglify can't convert ES6. Caused, it took several days to solve, but the ability is limited, can only wait.

gberaudo:
Hopefully I will be able to release a new version next week.

felix9ia:
Hard work! Thank you for your contribution to open source!

tlepkowski:
Yes, thank you for your hard work! I'm not sure if it's related, but I'm getting an assert error using the old-fashioned technique in the latest release.  I setup the example locally, worked fine.  Then plugged the same ol-cesium into my app and got the following error when **new olcs.OLCesium({map:ol2d})** :
`TypeError: invalid 'instanceof' operand type assertInstanceof asserts.js:6 createSingleLayerCounterparts VectorSynchronizer.js:106:4 _loop AbstractSynchronizer.js:127:24 addLayers_ AbstractSynchronizer.js:100:28 synchronize AbstractSynchronizer.js:77:4`
On client console.  Somehow other things (other OL layers, etc) in my real-world app have interfered with the asserts.  Let's see how this 636 fix works out.  Thank you!

PR #638, opened 2018-08-30
Internalize asserts

gberaudo:
Fixes: #636.


gberaudo:
There is a glitch with Travis / github interaction. Build are passing on Travis so I will merge.

PR #645, opened 2018-09-11
Uglify lib and examples

gberaudo:


gberaudo:
It is holiday time and there is nothing non-consensual about this change. I tested locally opening the examples (including the oldfashioned one) so merging it now.

PR #653, opened 2018-09-24
Correctly map icon anchor from IconStyle to Billboard

samuel-girard:
Hello,
By default in ol-cesium, icons are positioned differently in 2D (default is the anchor in the center of the icon) and in 3D (icon is bottom-aligned).
This fixes this problem as it calculates Cesium Billboard pixel offset from OpenLayer anchor position.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
 # Changelog
 
+# v 2.5
+
+* Breaking changes
+  * Icons are now anchored as specified in the OpenLayers style.
+
+* Changes
+  * CanvasPattern and CanvasGradient styles are now handled for polygons.
+
 # v 2.4 - 2018-10-01
 
 * Changes
--- a/examples/icon-position.html
+++ b/examples/icon-position.html
@@ -1,0 +1,23 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE HTML>
+<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
+  <head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+    <meta name="robots" content="index, all" />
+    <title>olcesium vectors example</title>
+    <link rel="stylesheet" href="../node_modules/ol/ol.css" type="text/css"></link>
+    <style>.cesium-credit-textContainer { display: inline-block; font-size: 50%; line-height: 100%;}</style>
+  </head>
+  <body>
+    <div id="map2d" style="width:600px;height:400px;float:left;"></div>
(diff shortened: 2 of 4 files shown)


gberaudo:
@samuel-girard, see lint errors:
/home/gberaudo/dev/ol3-cesium/src/olcs/FeatureConverter.js
  693:1  error  Expected indentation of 12 spaces but found 10  indent
  694:1  error  Expected indentation of 12 spaces but found 10  indent

samuel-girard:
Hi @gberaudo 
Sorry, I didn't check that before pushing. It is fixed now.
Thanks.

samuel-girard:
I proposed this fix because there is a difference on how icons are displayed in 2D and 3D.  
For example, in the vectors demo, you can see a difference:
image image
The circle is centered in the corner of the yellow square on OpenLayers map, but not in Cesium.

With this fix, the icon would be at the same place in 2D and 3D, which is, in my opinion, very important.
Also, both in OpenLayers (via `IconStyle`) and Cesium (via `BillBoard`), the default is to display the icon centered.

I understand what you are saying though, but if it is needed to display an icon over the ground, it is still possible to do it with `anchor` and `anchorOrigin` settings of OpenLayers `StyleIcon` (the same vectors demo displays correctly the icon image after the fix).

samuel-girard:
Hi @gberaudo 
I just added a simple example to show how to create an icon with this change:
- By default the icon is centered on the position:
```
new olStyleIcon({
  src: 'data/icon.png',
})
```
- Parameter `anchor: [0.5, 1]` will show the icon as they are now displayed in Cesium (anchor at the bottom center of the image)
```
new olStyleIcon({
  anchor: [0.5, 1],
  src: 'data/icon.png',
})
```

As it can be seen in the example, both OpenLayers and Cesium have the same behaviour for icon positioning, which was my original intention for this PR (also check how it corrects some icons positioning in the vectors example).

I still think that this should be included in ol-cesium. What do you think about it?

gberaudo:
@samuel-girard, thanks LGTM. Could you rebase, squash commits and update the changelog?

```diff
diff --git a/CHANGES.md b/CHANGES.md
index dd872e7..a82b041 100644
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
 # Changelog
 
+# v 2.5
+
+* Breaking changes
+  * Icons are now anchored as specified in the OpenLayers style.
+
+* Changes
+  * CanvasPattern and CanvasGradient styles are now handled for polygons.
+
 # v 2.4 - 2018-10-01
 
 * Changes
```

samuel-girard:
@gberaudo : thank you for your responsiveness

PR #660, opened 2018-09-28
CanvasPattern and CanvasGradient as polygon surface style

samuel-girard:
Hi,

When a surface style (`ol.style.Fill`) is defined as a CanvasPattern or CanvasGradient (as defined in the `ColorLike` type) in OpenLayers for a polygon, Cesium fails to display it.  
This change fixes this problem.

### DIFF
--- a/examples/fillstyle.html
+++ b/examples/fillstyle.html
@@ -13,6 +13,8 @@
     <div id="map2d" style="width:600px;height:400px;float:left;"></div>
     <div id="map3d" style="width:600px;height:400px;float:left;position:relative;"></div>
     <div><input id="enable" type="button" value="Enable/disable Cesium" /></div>
+    <div><input type="button" value="Toggle clampToGround mode"
+      onclick="javascript:toggleClampToGround(); ol3d.getAutoRenderLoop().restartRenderLoop()" /></div>
     <script src="inject_ol_cesium.js"></script>
   </body>
 </html>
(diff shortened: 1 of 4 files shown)


samuel-girard:
I just added an example to better show the update.  
The only problem I see is when we zoom in/out: in OpenLayers the canvas pattern is automatically repeated according to the resolution of the screen, but in Cesium the texture is static.  
Any idea on how to simulate similar result in Cesium than in OpenLayers?

gberaudo:
@samuel-girard,

Thanks for contributing this feature. Do you think this functionality is still useful despite the limitation you mentioned?

It is possible  to change the appearance repeat value: with new Cartesian(2, 2) the density of symbols become higher (with some artefacts). But I am not aware of a "resolution" dependent mechanism in Cesium (you should try asking the Cesium forum).

You should also handle ground primitives (see altitudeMode = 'clampToGround') and the code already present in OL-Cesium for handling it. The information is taken from the feature or the layer.

samuel-girard:
Thank you @gberaudo for your feedback.  
I updated the branch to also handle ground primitive as you suggested.  

To answer your question, yes I think this is still useful, because the current status is that ol-cesium does not recognize fill styles defined as CanvasGradient and CanvasPattern and raises errors when it happens.

I will later have a look on Cesium side if it is possible to have a more resolution-dependent solution, but you can merge this now if you don't see any other issue.

examples/fillstyle.html
```diff
@@ -0,0 +1,21 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<?xml version='1.0' encoding='UTF-8'?>
```
gberaudo: This line is duplicated.

PR #664, opened 2018-10-02
Minimalistic text over icon support

gberaudo:


gberaudo:
image

Otherwise the label on top of the OpenLayers logo is below the logo.

PR #670, opened 2018-10-24
calculate height for the SynchronizedOverlay using the globe.getHeight function

jbo023:
when cesium is configured with a terrain, the synchronized overlay with height 0 seems to be under the terrain, and changes the position depending on the viewpoint. 

This pull request fixes the problem and requests the height from the globe


gberaudo:
Thanks @jbo023.

PR #677, opened 2018-11-13
fix overlaysynchronizer 

jbo023:
Hi, 
during the ES6 Module updates this bug happend, but it only failed for overlays created before the synchronizer has been updated, so it didn't show in the overlay example.

This fixed the addOverlay function


gberaudo:
Thanks @jbo023.

PR #682, opened 2018-12-03
fix: update Webpack

samuel-girard:
Hello,  

The build is failing if uglifyjs Webpack dependency is not installed.  
This change fixes the error when trying to build the library.

### DIFF
--- a/buildtools/webpack.prod.js
+++ b/buildtools/webpack.prod.js
@@ -1,5 +1,5 @@
 const webpack = require('webpack');
-const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
+const TerserPlugin = require('terser-webpack-plugin');
 
 module.exports = {
   mode: 'production',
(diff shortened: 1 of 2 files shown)


samuel-girard:
Hi @gberaudo  
The build still fails after adding UglifyJs dependency. The first error I get is:  
```
ERROR in bing.3c269e30300df9237b0d.js from UglifyJs
Unexpected token: keyword (const) [./examples/bing.js:4,0][bing.3c269e30300df9237b0d.js:179,0]
```
It seems like it is not expecting ES6 ("const" keyword) at this point.  
If you succeeded in publishing version 2.5.0, what is your environment setup for the build? Do you have extra configuration for UglifyJs?  
Thanks

gberaudo:
I don't know why it fails. On my machine everything is passing.
See also webpack 4.20.2 package.json.  "uglifyjs-webpack-plugin": "^1.2.4" is already a dependency.

samuel-girard:
Ok, I see what is going on: uglifyjs has been removed from Webpack from version 4.26 (see https://github.com/webpack/webpack/commit/311a7285d36b38bada46102967c431e93ff48a89), and replaced with Terser.  
I updated this PR with the migration of Webpack instead, which fixes the uglifyjs problem I faced when the library is built.

gberaudo:
Thanks @samuel-girard  for the explanations and for committing a a fix.

PR #684, opened 2018-12-05
Synchronize interactions when in 3d mode

samuel-girard:
Hello,  

I got an issue in my application when I tried to remove a draw interaction on the map while 3D mode was enabled.  
I can see that when switching to 3D, interactions are temporarily kept into a `pausedInteractions` list then re-applied when switching back to 2D.  
This change allows to update pausedInteractions array when interactions are added or removed from the map when 3D mode is enabled.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 * Changes
   * Add new build as UMD
+  * Synchronize map interactions when 3D is enabled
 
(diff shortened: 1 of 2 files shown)


samuel-girard:
I don't like hacking with OpenLayers either, but like you I don't see any other easy way...

Issue #683, opened 2018-12-04
Build incompatible with modern frameworks


samuel-girard:
Hello,  

I have some issues with the current build of ol-cesium, when using it inside Angular application.  
My problem is similar with what I found in other open issues:
- the current setup makes that my application will use the entry point `src/index.library.js` (defined in the "module" key of package.json). The problem here happens in Internet Explorer 11, as it has no idea about ES2015 (console raises errors on `class` keyword): same problem than the issue opened here: https://github.com/openlayers/ol-cesium/issues/649
- i cannot used either `dist/olcesium.js` as it is gives the same error than https://github.com/openlayers/ol-cesium/issues/642

I have found a setup that makes me able to use the library perfectly fine in my application: https://github.com/samuel-girard/ol-cesium/commit/0702549033f2f5495bcfc6bf17c3641fc531901a  
But I don't know if this config is fine for other people using ol-cesium.

I cannot use `olcs` package, because Angular does not let me customize Webpack setup, so the only solution is to provide a library that is already built.

@gberaudo, what is your opinion on:
- either change the build with the modification I made
- either if this breaks other application setups, create another dist that would be more Angular-compatible. In this case, the new dist can still be built in the same `ol-cesium` npm package (for example, Angular libraries export multiple builds, with "esm5", "esm2015" keys in package.json to differentiate them.

gberaudo:
@samuel-girard, yes it will break the "legacy build" use case.
For the legacy build to work the imports to OpenLayers must be rewritten. For example `import .. 'ol/source/OSM.js'` must become `window.ol.source.OSM`.

It is unfortunate the JS world is becoming even more complicate than before. With more and more tricky/legacy cases instead of converging to the ES6 standard.

The OL-Cesium library is small and building it is fast. So I am OK with adding a new "legacy" build to handle your use case. Would you be able to create a PR for it?

gberaudo:
For the context and current workaround in Angular cli: https://github.com/angular/angular-cli/issues/10618.

samuel-girard:
Thanks for the info.  
I do agree that JS library builds are getting more complicated if you have to consider all possible usages.

I will try to get back soon with a PR that solve the integration with Angular.

PR #685, opened 2018-12-05
UMD build (for Angular projects using ol-cesium)

samuel-girard:
Hi @gberaudo  

Regarding https://github.com/openlayers/ol-cesium/issues/683, here is a change that makes ol-cesium works nicely inside Angular.  
I created a new build `olcesium.umd.js` which is pointed by the `module` property in the package.json  
It works well so far, and should not have impact on other projects as the previous build is not modified.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # Changelog
 
+# v 2.6
+
+* Changes
+  * Add new build as UMD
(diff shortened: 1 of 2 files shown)


gberaudo:
Please also add a note to the CHANGES.md and update the README.md.

PR #691, opened 2019-01-25
Exclude vector tile layers from synchronization

mschulz-716:
Ignore OL vectore tile layers during synchronization
since they are not supported currently and would cause
an error otherwise


gberaudo:
Thanks @mschulz-716.

PR #692, opened 2019-01-30
Convert olcsx to typedef

sbrunner:


gberaudo:
What about the documentation?

sbrunner:
Updated using typedoc to generate the documentation :-)

PR #712, opened 2019-09-17
Fix conversion of icon anchor to pixelOffset

Flowm:
In Cesium the offset of an entity is specified in pixels from it's center by
default. Openlayers uses the top left position as origin.
While the origin is already converted since #653, this does not consider any
scaling of the image that happens in Cesium at a later stage through the scale
parameter. With a scale < 1 this can result in images that are off by multiple
times their own size.

Multiplying the pixel offset with the scale of the image resolves the issue.

The icon-position example can be used to demonstrate this issue when a scale
parameter is added to the first icon.
```
diff --git a/examples/icon-position.js b/examples/icon-position.js
index 62fb17a..362e42b 100644
--- a/examples/icon-position.js
+++ b/examples/icon-position.js
@@ -28,6 +28,7 @@ icon1Feature.setStyle(new olStyleStyle({
   image: new olStyleIcon(/** @type {olx.style.IconOptions} */ ({
     anchor: [0.5, 1],
     src: 'data/icon.png',
+    scale: 0.2,
   })),
   text: new olStyleText({
     text: 'Icon with anchor on the bottom center',
```

Before:
image

After:
image

### DIFF
--- a/src/olcs/FeatureConverter.js
+++ b/src/olcs/FeatureConverter.js
@@ -714,13 +714,14 @@ class FeatureConverter {
         color = new Cesium.Color(1.0, 1.0, 1.0, opacity);
       }
 
+      const scale = imageStyle.getScale();
       const heightReference = this.getHeightReference(layer, feature, olGeometry);
 
       const bbOptions = /** @type {Cesium.optionsBillboardCollectionAdd} */ ({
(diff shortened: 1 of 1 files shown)


gberaudo:
Thanks @Flowm for your contribution.

Issue #741, opened 2020-03-02
Custom matrix size for WMTS


remiHau:
Hello,

I'm working with openlayers 6.01 and ol-cesium 2.10.0 and I'm currently trying to consume a wmts layer which has a matrix of 2x1 on level 0.

My layer is not tiled correctly in Cesium. 
I saw in the code that OlImageryProvider convert all the layer image in a `tilingScheme` with default values for the number of tile at level 0 and I think my problem come from that.
I pass the matrixSize as a parameter of my wmts source and used it in OlImageryProvider in a fork of ol-cesium and it seems to have corrected my bug.

I was wondering if there was a method to modify the number of tile at level 0 for a layer wmts or if it should be calculated from another parameter from the layer (which may be missing).

If that's not the case, do you think it's possible to add a PR to modify OlImageryProvider to create the `tilingScheme` from the parameters of the layers? 
The `matrixSize` is not a parameter of the `tileGrid` so I am not sure how to get this information, maybe from the attribute `sizes`.

On a maybe related note, I saw that the method requestImage of OlImageryProvider request the next zoom level if the projection of the layer is EPSG:4326. I doesn't seem to be the case with my layer wmts and I don't understand why it does that.

Thank you in advance for your help.

gberaudo:
@remiHau, I think the "+1" zoom difference was due to Cesium using a matrix of 2x1 on level 0, which is not usual. If you can come up with a PR it will help me understand what the problem is.

remiHau:
Hello, 
Sorry for the delay, I had trouble to reproduce my error outside my work environment.

So I forked ol-cesium and I create a new example to show you the issue.
You can find it here : https://github.com/remiHau/ol-cesium/blob/master/examples/wmts-issue.js

Below, I took a screenshot of my issue :
<img width="181" alt="example-bug" src="https://user-images.githubusercontent.com/12863346/76980609-ad506180-6939-11ea-9bc4-830a3d51cca1.png">

As you can see, the globe contains only a quarter of the map but it renders correctly in openlayers.
I think the issue is this layer is using a custom matrix and it conflicts with the "+1" correction.

I am gonna work on a PR to correct the issue but I was wondering if you had an idea of how to define the matrix of the tileGrid when the layer is created in Cesium.

Thank you for your help.

mirosvanek:
I'm also interested in using ol6

remiHau:
[...]

PR #744, opened 2020-03-20
Manage non-standard tileGrid WMTS

remiHau:
Correct a bug for non-standard tileGrid for WMTS layer.
Fixes :#741

### DIFF
--- a/src/olcs/core/OLImageryProvider.js
+++ b/src/olcs/core/OLImageryProvider.js
@@ -5,7 +5,6 @@ import {get as getProjection} from 'ol/proj.js';
 import olcsUtil from '../util.js';
 import {Tile as TileSource} from 'ol/source.js';
 
-
 const olUseNewCoordinates = (function() {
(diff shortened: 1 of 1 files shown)


gberaudo:
@remiHau, in order to know the number of tiles at zoom level 0, you might be able to use the  [forEachTileCoord](https://openlayers.org/en/latest/apidoc/module-ol_tilegrid_TileGrid-TileGrid.html#forEachTileCoord) method which is part of the OL API:

```js
let count = 0;
tilegrid.forEachTileCoord(the_projection_extent, the_zoom_level, function() {count += 1;}):
shouldUseNextLevel = ... count === 1;
```
With this method you may also determine the maxX, maxY if needed.

remiHau:
Thank you, that was very helpful to determine maxX and maxY. I did'nt create the 'count' variable because I have already the maxX and maxY but I can add it if you think it is more readable.

gberaudo:
Thanks @remiHau.

PR #767, opened 2020-06-26
Update dependencies

gberaudo:


gberaudo:
@fredj, I have added a "peerDependency" to the package.json, since the lib is intented to be used with OpenLayers and CesiumJs but do not provide them.

Not sure it is correct though since people could prefer using their forks of OpenLayers and CesiumJS in their apps and do an alias with webpack or other builder.

PR #783, opened 2020-08-19
Update usage guidelines for Angular

bampakoa:
Change the way that the library is imported into Angular applications.


gberaudo:
Thanks for spotting this.

PR #823, opened 2020-09-30
Use LRU caching strategy

gberaudo:
Cache size is set to the arbitrary value 50.


src/olcs/MVTImageryProvider.js
```diff
@@ -29,14 +30,16 @@ export default class MVTImageryProvider {
...
     this.hasAlphaChannel = true;
-    this.cache_ = {};
     this.styleFunction_ = options.styleFunction || (() => styles);
     this.projection_ = getProjection('EPSG:3857');
     this.emptyCanvas_ = document.createElement('canvas');
     this.emptyCanvas_.width = 1;
     this.emptyCanvas_.height = 1;
-    this.featuresCache = options.featuresCache || {};
     this.tileRectangle_ = new Cesium.Rectangle();
+    const cacheSize = options.cacheSize ? options.cacheSize : 50;
```
fredj, line 39: Could be useful if we want to test with cacheSize of 0
```suggestion
    const cacheSize = options.cacheSize !== undefined ? options.cacheSize : 50;
```

fredj (approved):

PR #991, opened 2022-05-04
Compare projections by code

mschulz-716:
We are maintaining layers in our NgRx Store as part of the application state. Therefore we need to deep copy the layers including the projections when updating the state. Comparing projection objects therefore does not work in our case and we would be happy if the projection comparison could be switched to comparing them by their code.


gberaudo (approved):

gberaudo:
Thanks for the PR @mschulz-716.

PR #1004, opened 2022-10-10
Update dependencies

gberaudo:

PR #1005, opened 2022-10-24
Fix #998 - ol layer vectorImage is not rendered

### LINKED ISSUE #998: vectorImageLayer are not rendered in the 3d map
now that render mode: 'image' is deprecated , I am using VectorImageLayer  to display vector layers with hundreds of features 
However  this type of layer is not shown in the 3d map.
will olcs support vectorImageLayer or is is there a property of ol/layer/vector that I can use instead of rendermode:'image'?

or4inveni:
#998 simple fix for the issue 998 like discussed with @gberaudo 

Warning: as the current latest cesium version 1.98.0 is breaking the whole ol-cesium library this commit limits the cesium version to 1.97.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "@mapbox/geojsonhint": "^3.0.1",
     "@types/webpack": "^5.28.0",
     "babel-loader": "8.2.2",
-    "cesium": "^1.81.0",
+    "cesium": ">=1.81.0 <=1.97.0",
     "copy-webpack-plugin": "^9.0.0",
     "cross-env": "7.0.3",
     "eslint": "^7.26.0",
--- a/src/olcs/VectorSynchronizer.js
+++ b/src/olcs/VectorSynchronizer.js
@@ -10,6 +10,7 @@ import olLayerVector from 'ol/layer/Vector.js';
 import olLayerVectorTile from 'ol/layer/VectorTile.js';
 import olcsAbstractSynchronizer from './AbstractSynchronizer.js';
 import olcsFeatureConverter from './FeatureConverter.js';
+import olLayerVectorImage from 'ol/layer/VectorImage.js';
 
 class VectorSynchronizer extends olcsAbstractSynchronizer {
   /**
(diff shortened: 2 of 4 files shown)


src/olcs/VectorSynchronizer.js
```diff
@@ -99,8 +100,16 @@ class VectorSynchronizer extends olcsAbstractSynchronizer {
...
   createSingleLayerCounterparts(olLayerWithParents) {
-    const olLayer = olLayerWithParents.layer;
-    if (!(olLayer instanceof olLayerVector) || olLayer instanceof olLayerVectorTile) {
+    let olLayer = olLayerWithParents.layer;
+
+    if(olLayerWithParents.layer instanceof olLayerVectorImage){
+      let convertedVectorLayer = new olLayerVector({source: null});
+      convertedVectorLayer.setStyle(olLayerWithParents.layer.getStyle())
+      convertedVectorLayer.setSource(olLayerWithParents.layer.getSource())
+      olLayer = convertedVectorLayer;
```
gberaudo, line 109: Please directly pass source and style as constructor parameter.
Notably, see https://openlayers.org/en/latest/apidoc/module-ol_layer_Vector-VectorLayer.html

package.json
```diff
@@ -31,7 +31,7 @@
     "@mapbox/geojsonhint": "^3.0.1",
     "@types/webpack": "^5.28.0",
     "babel-loader": "8.2.2",
-    "cesium": "^1.81.0",
+    "cesium": ">=1.81.0 <=1.97.0",
```
gberaudo, line 34: No, see https://github.com/openlayers/ol-cesium/pull/1004.

package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "ol-cesium",
-  "version": "2.13.1",
+  "name": "@inveni/ol-cesium",
+  "version": "0.00.1",
```
gberaudo, line 3: Don't change this.

gberaudo (changes requested):
Thanks for your PR. Could you please address my comments?

src/olcs/VectorSynchronizer.js
or4inveni: will fix it next week

ninichuuh:
Hi is this fixed? Cause i still have problems showing image vector layers in 3d?

Issue #1006, opened 2022-10-25
const exports = {} fails on bundling with webpack and esbuild


geodatenanwendungenhamburg:
We changed the bundling with webpack and replaced babel-loader with esbuild-loader.
This raises in an error in olcs:

> stderr: 'Error: Transform failed with 2 errors:\n' +
    'js/masterportal.js:30161:6: ERROR: The symbol "exports" has already been declared\n' +
    'js/masterportal.js:45327:6: ERROR: The symbol "exports" has already been declared\n' +
    '    at failureErrorWithLog (C:\\foo\\baa\\masterportal\\node_modules\\esbuild\\lib\\main.js:1566:15)\n' +
    '    at C:\\foo\\baa\\masterportal\\node_modules\\esbuild\\lib\\main.js:805:29\n' +
    '    at responseCallbacks.<computed> (C:\\foo\\baa\\masterportal\\node_modules\\esbuild\\lib\\main.js:671:9)\n' +
    '    at handleIncomingPacket (C:\\foo\\baa\\masterportal\\node_modules\\esbuild\\lib\\main.js:726:9)\n' +
    '    at Socket.readFromStdout (C:\\foo\\baa\\masterportal\\node_modules\\esbuild\\lib\\main.js:647:7)\n' +
    '    at Socket.emit (node:events:527:28)\n' +
    '    at addChunk (node:internal/streams/readable:315:12)\n' +
    '    at readableAddChunk (node:internal/streams/readable:289:9)\n' +
    '    at Socket.Readable.push (node:internal/streams/readable:228:10)\n' +
    '    at Pipe.onStreamRead (node:internal/stream_base_commons:190:23)\n'

The reason is:

`const exports = {};`

in olcs/util.js and in olcs/core.js

Can you rename the variable exports?

At the moment we are renaming the originating code during the build process.

gberaudo:
Hi @geodatenanwendungenhamburg. Indeed these `exports` variables are the remaining of pre-ES6 module system.
These should all be removed from the code (we should use standard named exports).

Would you want to contribute a PR to fix that?

eblondel:
@gberaudo I suppose #1012 is the same issue (on Parcel). I might have to time to contribute to a PR, but i'd need more guidance. Do you expect to create classes for core and utils? as for the other js files here. 
Thanks in advance

gberaudo:
I @eblondel , thanks for proposing your help.
You simply need to create named exports/imports. Like following example:

```js
/**
[...]

PR #1013, opened 2022-12-05
Fix for #1006 and #1012

### LINKED ISSUE #1012: Bundling with parcel ^2.8.0 throws errors, because of 2x const exports = {};
If parcel ^2.8.0 is used, an error occurs when bundling, because in olcs 2x `const exports = {};` is declared, which leads to the following error in the built project:

`Uncaught SyntaxError: Identifier 'exports' has already been declared (at index.c9c984ed.js:89116:7)`

I was able to find the places in olcs where this is done:
https://github.com/openlayers/ol-cesium/blob/master/src/olcs/core.js (line 20)
https://github.com/openlayers/ol-cesium/blob/master/src/olcs/util.js olcs\util.js (line 4)

Would it be possible to replace the `const `with `var `or solve the problem differently?

eblondel:
@gberaudo Here's a proposal of base patch to fix #1012  and probably #1006 

The code suggests minimal changes by handling core.js `exports` as class with static methods; and util.js with a simple class for which we export an instance of it as const. In that way, imho we have simplified imports in the underlying js files.

I've successfully tested the code in this test repo: https://github.com/eblondel/troubleshooting-parcel-olcs and its integration in the map viewer I develop at [OpenFairViewer](https://github.com/eblondel/OpenFairViewer) that includes ol-cesium

### DIFF
--- a/src/olcs/Camera.js
+++ b/src/olcs/Camera.js
@@ -5,7 +5,7 @@
 import {unByKey as olObservableUnByKey} from 'ol/Observable.js';
 import {toRadians, toDegrees} from './math.js';
 import {getTransform} from 'ol/proj.js';
-import olcsCore, {calcDistanceForResolution, calcResolutionForDistance} from './core.js';
+import {olcsCore, calcDistanceForResolution, calcResolutionForDistance} from './core.js';
 
 class Camera {
   /**
--- a/src/olcs/FeatureConverter.js
+++ b/src/olcs/FeatureConverter.js
@@ -8,9 +8,9 @@ import olSourceCluster from 'ol/source/Cluster.js';
 import {circular as olCreateCircularPolygon} from 'ol/geom/Polygon.js';
 import {boundingExtent, getCenter} from 'ol/extent.js';
 import olGeomSimpleGeometry from 'ol/geom/SimpleGeometry.js';
-import olcsCore from './core.js';
+import {olcsCore} from './core.js';
 import olcsCoreVectorLayerCounterpart from './core/VectorLayerCounterpart.js';
-import olcsUtil, {getUid, isGroundPolylinePrimitiveSupported} from './util.js';
+import {olcsUtil, getUid, isGroundPolylinePrimitiveSupported} from './util.js';
 
 
 /**
--- a/src/olcs/OLCesium.js
+++ b/src/olcs/OLCesium.js
@@ -3,8 +3,8 @@
  */
 import olGeomPoint from 'ol/geom/Point.js';
 import {getTransform} from 'ol/proj.js';
-import olcsUtil from './util.js';
-import olcsCore from './core.js';
+import {olcsUtil} from './util.js';
+import {olcsCore} from './core.js';
 import olcsAutoRenderLoop from './AutoRenderLoop.js';
 import olcsCamera from './Camera.js';
 import olcsRasterSynchronizer from './RasterSynchronizer.js';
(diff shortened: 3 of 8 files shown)


gberaudo:
Hi @eblondel, thanks for the PR and for sharing a link to your application.

I really think we should use named exports here instead of wrapping all as static functions attached to a class.
Because:
- it is artificial: there is no need for a host object in JS;
- it probably prevents proper treeshaking, which is a must have for libraries.

eblondel:
For the core.js indeed i created a static class. Indeed, this could be separate functions (but same logic would have applied for the initial 'exports' object). So the code i pushed is the same as the current code, but formalized as actual class, and is maybe more easily maintainable for its import in the other Js rather than importing all core functions needed.

For the util.js, i'm not sure how you want it, because the util.js `exports` object has some private fields, so in its structure sounds an object-oriented approach, although not formalized as such.

Let me know

gberaudo:
Hi @eblondel,

`OL-Cesium` was created before the ES6 modules specifications. If `named exports` had existed, we would have used them. So indeed this work is modernization of the codebase.

For `util.js`, these private properties should be made global to the file:
```js
/**
 * @type {boolean|undefined}
 * @private
 */
exports.supportsImageRenderingPixelatedResult_ = undefined;
```
-> should become
```js
/**
 * @type {boolean|undefined}
 */
let supportsImageRenderingPixelatedResult_ = undefined;
```

eblondel:
@gberaudo done. Sorry i had some wrong push initially hence the various fix commits. Let me know if it's ok, or if you want me to send a new clean single-commit pull request.

gberaudo:
Hi @eblondel, thanks for your work.
Could you run `make lint` and fix the reported errors? Then can you check the examples are still working properly?

gberaudo:
Hi @eblondel,

I have never tried on Windows but some in the community did.
Are you using [Windows Subsystem for Linux](https://learn.microsoft.com/en-us/windows/wsl/install)?
If not, it is probably worth to try.

Anyway, here is the part of the Makefile related to linting:
```make
.build/eslint.timestamp: $(SRC_JS_FILES) $(EXAMPLES_JS_FILES) $(WEBPACK_CONFIG_FILES) .build/node_modules.timestamp
        TARGET=examples ./node_modules/.bin/eslint $(filter-out .build/node_modules.timestamp, $^)
        touch $@
```

You can do a manual lint by passing the list of JS from the sources and examples directory to `eslint`.
Give it a try and let me know if you have a problem.

eblondel:
Thanks @gberaudo  for the advice
i've finally got sometimes to configure WSL. I could run `make lint` from there very easily. However before that I had to configure `eslint --init`.

* If I choose the minimal option (vheck syntax), I get an error that has nothing to deal with my pull request:

```
/mnt/d/Documents/DEV/www/ol-cesium/webpack.config.js
  1:1  error  Definition for rule 'import/no-unresolved' was not found  import/no-unresolved
✖ 1 problem (1 error, 0 warnings)
```
On this, it doesn't seem I did something wrong on the modified files.

* If I choose intermediate option to check syntax and detect problems, I get hundreds of errors, but here as well many dealing with files that I didn't touch at all..
* If I choose full check (including code styling), here i've thousands of errors, essentially dealing with code styling, and all files of the repo

Let me know

gberaudo:
Hi @eblondel. thank you for your effort on this PR. I was finally able to take the time to look again into it thanks to the code sprint that is happening now.

PR #1021, opened 2023-04-25
Fix link in README.md

louwers:
I got an error about an expired SSL certificate and got redirected to this URL.


fredj:
Thanks @louwers !

PR #1028, opened 2023-04-26
Remove workaround to prevent camera from going underground

fredj:


gberaudo (approved):
Not sure this is still required or not.
Let's remove it and revisit if someone complains about it.

PR #1029, opened 2023-04-26
Depend on older Cesium version

gberaudo:
Latest Cesium breaks most examples.
With this commit we use the previous version while work is in progress fixing the issue.


fredj:
examples fixed with #1030

PR #1044, opened 2023-04-27
Add shadows and light to the scene

pedroslvieira:


src/olcs/FeatureConverter.js
```diff
@@ -176,9 +176,20 @@ class FeatureConverter {
...
+          translucent: color.alpha !== 1,
+          fabric: {
+            type: 'Color',
+            uniforms: {
+              color,
+            }
+          }
+        })
+      });
+      primitive.shadows = 1;
```
gberaudo, line 191: Casting shadows have performance implications.
So I would have it opt-in (like in CesiumJS).

Let's set it if one of the following properties is set:
- layer.get('olcs_shadows')
- feature.get('olcs_shadows')

src/olcs/FeatureConverter.js
```diff
@@ -176,9 +176,26 @@ class FeatureConverter {
...
+            }
+          }
+        })
+      });
+      if (feature.get('olcs_shadows') || layer.get('olcs_shadows')) {
+        primitive.shadows = 1;
+        if (!this.scene.shadowMap.enabled) {
+          this.scene.shadowMap.enabled = true;
+          this.scene.globe.enableLighting = true;
+        }
```
gberaudo, line 196: I understand that shadows will not work without this block.
But I don't think it is correct to enable these scene options automatically behind the scene.

You should:
- document `olcs_shadows` somewhere and explain there that you need these options enabled;
- you should enable them in the example and add a  comment there too.

gberaudo (approved):

PR #1069, opened 2023-07-06
Move limitCameraToBoundingSphere to core

fredj:


gberaudo (approved):
OK for me but you should keep the `limitCameraToBoundingSphere` method (call core from there).

src/olcs/core.ts
```diff
@@ -774,3 +775,32 @@ export function calcResolutionForDistance(distance: number, latitude: number, sc
...
+ */
+export function limitCameraToBoundingSphere(camera: Camera, boundingSphere: BoundingSphere, ratio: (height: number) => number): () => void {
+  let blockLimiter = false;
+  return function () {
+    if (!blockLimiter) {
+      const position = camera.position;
+      const carto = Cesium.Cartographic.fromCartesian(position);
+      if (Cesium.Cartesian3.distance(boundingSphere.center, position) > boundingSphere.radius * ratio(carto.height)) {
+        // @ts-ignore TS2339: FIXME, there is no flying property in Camera
+        const currentlyFlying = camera.flying;
```
fredj, line 791: @gberaudo do you remember why we check this property? I could find anything in cesium or ol-cesium about this flying property

gberaudo: @fredj, we need to disable the camera limits when flying because the fly path can go out of the bounding sphere.
I used this internal `flying` property which was set by Cesium. I don't know if it is still the case.

PR #1076, opened 2023-07-21
Add client side printing functions

gberaudo:
- display a mask;
- take a screenshot.


src/olcs/print/computeRectangle.ts
```diff
@@ -0,0 +1,29 @@
+export function computeRectangle(canvas: HTMLCanvasElement, tw: number, th: number) {
```
fredj, line 1: Return type is `ScreenshotOptions` ?

fredj (approved):

Issue #1065, opened 2023-06-30
Pinch Interactions not working properly in 3D


geodatenanwendungenhamburg:
Hey everybody,

we noticed that pinch interactions to rotate or to zoom are not working properly on the recent android/chrome versions in 3D (safari as well). If you open https://openlayers.org/ol-cesium/examples/overlay.html for example, Rotating via pinch in cesium is being aborted after minimal rotating while interactions on the ol-side are working. We're not getting these effects with 'naked' cesiumapp like for example in https://cesium.com/learn/cesiumjs-learn/cesiumjs-quickstart/. So we assumed that this may relate to ol-cesium. Also to be seen in https://www.geoportal-hamburg.de/geo-online/?map=3d.

best regards
Markus

ismailsunni:
It seems it's similar to this issue #981 

and yes, I got the same issues.

gberaudo:
Hi @geodatenanwendungenhamburg, thanks for your precise feedback.

I can indeed reproduce it with chrome on android.
Firefox works correctly.

I agree with you @ismailsunni.

gberaudo:
Hi @geodatenanwendungenhamburg, the problem is linked to https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action#none

@ismailsunni is working on a fix.

geodatenanwendungenhamburg:
Oh, thanks for the investigation!

"@ismailsunni is working on a fix." - sounds great :-)

Issue #981, opened 2021-09-04
pinch zoom in 3D does not work on mobile devices


lukey78:
Apparently, the pinch zoom gesture does not work with OLCesium on mobile devices. 

You can check by opening an example, e.g. https://openlayers.org/ol-cesium/examples/kml.html, and then try to zoom with the zoom gesture.

Pinch zoom always zooms the whole web page, but not the 3D content. 

In "plain Cesium" this gesture works.
Is there a way to repair this behaviour?

Thanks.

gberaudo:
Hello @lukey78 thanks for the report. I tested this example on my smartphone, both in Firefox and chrome, and it works perfectly well. On which device/browser do you encounter this issue?

lukey78:
I tried this with an iPhone in Chrome and Safari, and also with an iPad in both browsers. Maybe it's a Apple only issue?

Please see the attached video: I opened the example, moved and pinched in the 2d map (working perfectly), then I switched to 3D mode and did the same -> pinch zooming zooms the whole web page but not the 3d content.

https://user-images.githubusercontent.com/2877837/132748284-15372b55-1563-431f-9465-c5517feba86a.mov

gberaudo:
Hi @lukey78, it may be related to this: https://stackoverflow.com/questions/37808180/disable-viewport-zooming-ios-10-safari.
In the examples we have no headers preventing zoom. I don't have an Apple device to test. Could you continue investigating this?

lukey78:
Hi @gberaudo - this MAY be related, but it doesn't explain why pinch zooming works in the 2D map but not in the 3D map. The example site uses the same headers. If it'd be a viewport issue then it should affect the 2D map, too. 

Does Ol-Cesium change the touch event listeners somehow?

gberaudo:
> Does Ol-Cesium change the touch event listeners somehow?

@lukey78, no special action on these events:
- there is the `stopOpenLayersEventsPropagation` option that prevent OpenLayers to receive events, but it is not enable by default;
- we are using the default controls from Cesium so the behaviour should be the same.

lukey78:
Strange. Do you know of any working example where someone has used ol-cesium in a mobile app and is able to zoom the 3D view on Apple devices?

gberaudo:
@lukey78, can you try https://map.geo.admin.ch/ ? Make sure to click the 3D button as Cesium is lazy loaded.

lukey78:
[...]

PR #1086, opened 2023-08-23
Disables touch action on canvas to fix pinch zoom problem

pedroslvieira:
Correction for issues [#1065](https://github.com/openlayers/ol-cesium/issues/1065) and  [#981](https://github.com/openlayers/ol-cesium/issues/981)


gberaudo:
Thanks @pedroslvieira for the fix.

PR #1089, opened 2023-08-25
Fix options passing to Cesium.MaterialAppearance

### LINKED ISSUE #1075: Selection example is broken
When selecting the country highlight width should change to 3 pixels.
There is no change curently in the live examples.

fredj:
fixes #1075


gberaudo:
Thanks @fredj !

PR #1109, opened 2023-09-21
fix: cesium container is not initially hidden

### LINKED ISSUE #1108: Container is not hidden initially
The DIV container that wraps the Cesium canvas and the ol-overlayer containers is not hidden initially.

I've discovered that the issue stems from a missing semicolon in the `fillArea` string, which affects the correct styling of the container.

image

I'm going to prepare a pull request.

MauroIT:
Fixes #1108


gberaudo (approved):

gberaudo:
Thanks @MauroIT.

PR #1125, opened 2023-11-14
perf(text-offset): Unidirectional offset

XiaofengZeng:
There is an OpenLayers vector layer `style` that has both `image` and `text` attributes, and the text cannot overlap with the image. So, it is necessary to set a one-way `offsetY` for the text. However, when synchronizing to OLCesium, the corresponding `Cesium.Label` instance was not offset. Therefore, I adjusted the relevant code for the offset attribute to support one-way offset for the text.


gberaudo (approved):

gberaudo:
Thanks again @XiaofengZeng .

PR #1131, opened 2023-12-11
perf(overlay): delete duplicated code.

XiaofengZeng:
When I was looking at the source code, I found that the registration event in SynchronizedOverlay was duplicated, and I think it should be removable.
position


gberaudo (approved):

gberaudo:
Thanks @XiaofengZeng .

PR #1168, opened 2024-02-19
Add export for OverlaySynchronizer in olcs.ts

luca-rath:
I want to pass a custom `createSynchronizers` function to the `OLCesium` constructor with the default `OverlaySynchronizer`, the default `VectorSynchronizer` and a custom `RasterSynchronizer`. I don't know why the `OverlaySynchronizer` isn't exported in the main file, if I import it from `olcs/src/olcs/OverlaySynchronizer`, there are typescript compilation errors, if I add the export to the `olcs.ts` file and import it from there, everything works just fine.


gberaudo (approved):

gberaudo:
Thanks @luca-rath.

PR #1177, opened 2024-03-12
Optional synchronisation optimization 3d to 2d

tschuege9008:
This allows to only synchronize from 3d to 2d after the cesium camera stopped moving. This saves a lot of unnecessary overhead, if the 2d map is not visible anyways.
The default behavior is to synchronize continuously as before.


gberaudo (approved):
Thanks!

Issue #666, opened 2018-10-12
ol.layer.Image doesn't work with custom imageLoadFunction


drnextgis:
I made a simple example with `ol.source.ImageWMS` and custom `imageLoadFunction`. In 2D view all works fine but if switch to 3D then layer is disappeared: http://jsfiddle.net/_DR_/zft7bg1w/

oeninghe-dataport:
We also have this problem in our project. The problem is that in https://github.com/openlayers/ol-cesium/blob/master/src/olcs/core.ts#L361 , an image WMS is converted iff it uses the default `imageLoadFunction`. A possible solution might be to use the `imageLoadFunction` as `tileLoadFunction`, as both usually use only the `getImage()` function of its corresponding first parameter. However, the types do not match, so there is some TypeScript magic needed

geodatenanwendungenhamburg:
Hendrik Oenings implements for the masterportal and we need the fix of this bug.
Can you estimate when it will be fixed?

gberaudo:
Hi @oeninghe-dataport  and @geodatenanwendungenhamburg , I have created a PR to change this code.
Indeed it is probably safe to always do the conversion.

gberaudo:
FYI I published version 3.20 with this change.

PR #1190, opened 2024-04-23
Convert all image WMS load functions

gberaudo:
Fixes: #666


gberaudo:
@fredj, what do you think about it?
Is it reasonably safe to bypass the ImageWrapper thing in that case?

PR #1197, opened 2024-05-23
fix: parcel serve target dist directory

pedroslvieira:
The examples were not working on dev environment after running `npm start` because there is a bug in parcel that is ignoring the distDir attribute in targets for `parcel serve`. I replaced with cli commands and added a `prestart` hook with the commands that were in `build-examples` to add parts that are not compiled by parcel (should we also investigate that?).

bug reports in parcel:
https://github.com/parcel-bundler/parcel/issues/8715
https://github.com/parcel-bundler/parcel/issues/8105
https://github.com/parcel-bundler/parcel/discussions/6778 

Also, should I just remove the targets object from the code?

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
     "lint": "eslint src buildtools examples",
     "checks": "npm run lint && npm run typecheck",
     "doc": "typedoc --name ol-cesium --excludeExternals --out apidoc --entryPoints src/olcs.ts --tsconfig ./tsconfig.json",
-    "start": "parcel serve --no-autoinstall --target examples",
+    "prestart": "rm -rf dist; mkdir -p dist/examples/node_modules/ol; mkdir -p dist/examples/node_modules/cesium/Build/; cp examples/inject_ol_cesium.js dist/examples/; cp node_modules/ol/ol.css dist/examples/node_modules/ol/; cp -R node_modules/cesium/Build/CesiumUnminified node_modules/cesium/Build/Cesium dist/examples/node_modules/cesium/Build/; cp -R examples/data dist/examples/",
+    "start": "parcel serve ./examples/index.html --dist-dir dist/examples --no-autoinstall",
     "serve-built-examples": "python3 -m http.server --directory dist 12345"
   },
   "targets": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -10,7 +10,8 @@
     "lint": "eslint src buildtools examples",
     "checks": "npm run lint && npm run typecheck",
     "doc": "typedoc --name ol-cesium --excludeExternals --out apidoc --entryPoints src/olcs.ts --tsconfig ./tsconfig.json",
-    "start": "parcel serve --no-autoinstall --target examples",
+    "prestart": "rm -rf dist; mkdir -p dist/examples/node_modules/ol; mkdir -p dist/examples/node_modules/cesium/Build/; cp examples/inject_ol_cesium.js dist/examples/; cp node_modules/ol/ol.css dist/examples/node_modules/ol/; cp -R node_modules/cesium/Build/CesiumUnminified node_modules/cesium/Build/Cesium dist/examples/node_modules/cesium/Build/; cp -R examples/data dist/examples/",
```
gberaudo, line 13: Instead of duplicating you can do like this:
```suggestion
    "copy-static-dist": "mkdir -p dist/examples/node_modules/ol; mkdir -p dist/examples/node_modules/cesium/Build/; cp examples/inject_ol_cesium.js dist/examples/; cp node_modules/ol/ol.css dist/examples/node_modules/ol/; cp -R node_modules/cesium/Build/CesiumUnminified node_modules/cesium/Build/Cesium dist/examples/node_modules/cesium/Build/; cp -R examples/data dist/examples/"
    "prestart": "rm -rf dist; npm run copy-static-dist",
    "build-examples": "rm -rf dist; parcel build && npm run copy-static-dist",
```

gberaudo (commented):
Thanks for the fix.

I think you should also remove the `ensureExamplesUrl` function.
Also, the `dist-dir` CLI parameter is supposed to be a fallback for targets without `distDir` so yes that property should be removed from the target.

package.json
pedroslvieira: added the `dist-dir` parameter to build-examples, it was taking it from the target

gberaudo:
Thanks again!

PR #1238, opened 2024-10-10
Fix missing .js extensions in imports

ejn:
I can't see any reason why the `.js` is missing on these two imports (as far as I can see they are the only ones), and the fact they are missing is causing webpack to choke in a downstream project.


gberaudo (approved):

gberaudo:
Indeed @ejn , thanks.

PR #1245, opened 2024-11-05
feature: Expose OLCesiumOptions

ejn:
Downstream projects find it helpful to be able to reuse the type definition of the constructor options for the main OLCesium object.

I also corrected the type of the `sceneOptions` option: The `canvas` and `scene3DOnly` options are set internally and so should not be passed as part of the `sceneOptions`. Previously `canvas` strictly had to be passed as it is defined as a non-optional member of `SceneOptions`.

### DIFF
--- a/src/olcs.ts
+++ b/src/olcs.ts
@@ -1,6 +1,8 @@
 import OLCesium from './olcs/OLCesium.js';
 export default OLCesium;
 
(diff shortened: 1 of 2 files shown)


gberaudo (approved):

gberaudo:
Looks sensible, thanks @ejn.

PR #1305, opened 2026-03-18
rotateAroundAxis Use nullish coalescing instead of defaultValue

mapeGeoinfo:
Use nullish coalescing instead of defaultValue for using rotateAroundAxis with newer cesium versions. defaultValue is deprecated since 1.128 and removed since 1.134


gberaudo:
Thanks @mapeGeoinfo, there is an extra file in the PR. I have opened a new PR to fix it.

PR #1337, opened 2026-06-26
Fix:#1338 the issue where the camera bounding sphere constraint is not actually taking effect

### LINKED ISSUE #1338: camera bounding sphere limit not applying
In core.ts, limitCameraToBoundingSphere() returns a closure that must be saved and executed by the caller to apply the constraint. However, the Manager method discards the returned value after calling it, meaning the camera is never actually constrained

xytczh:
In core.ts, limitCameraToBoundingSphere() returns a closure that must be saved and executed by the caller to apply the constraint. However, the Manager method discards the returned value after calling it, meaning the camera is never actually constrained.

Fixes #1338.

### DIFF
--- a/src/olcs/contrib/Manager.ts
+++ b/src/olcs/contrib/Manager.ts
@@ -162,11 +162,12 @@ export default class Manager extends Observable {
    */
   protected limitCameraToBoundingSphere() {
     const scene = this.ol3d.getCesiumScene();
-    limitCameraToBoundingSphere(
+    const limiter = limitCameraToBoundingSphere(
       scene.camera,
       this.boundingSphere_,
       this.limitCameraToBoundingSphereRatio,
     );
+    limiter();
   }
 
   /**
--- a/src/olcs/contrib/Manager.ts
+++ b/src/olcs/contrib/Manager.ts
@@ -162,12 +162,14 @@ export default class Manager extends Observable {
    */
   protected limitCameraToBoundingSphere() {
     const scene = this.ol3d.getCesiumScene();
-    const limiter = limitCameraToBoundingSphere(
-      scene.camera,
-      this.boundingSphere_,
-      this.limitCameraToBoundingSphereRatio,
-    );
-    limiter();
+    if(this.boundingSphere_){
+      const limiter = limitCameraToBoundingSphere(
+        scene.camera,
+        this.boundingSphere_,
+        this.limitCameraToBoundingSphereRatio,
+      );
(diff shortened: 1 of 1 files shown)


ahocevar:
Thanks — this does fix the real bug: the original code created the limiter closure but never invoked it, so the constraint never ran. The added `limiter()` call is what actually fixes it (the allocation itself is not strictly necessary — it could also have been called as an IIFE).

One concern though: since this method runs on every `postRender`, you now allocate a fresh closure each frame, which resets `blockLimiter` to `false` every time. That defeats the debounce the closure was designed for — re-triggering during a `flyToBoundingSphere` is then only prevented by the secondary `camera.flying` guard.

The factory is meant to be called once. Cleaner fix: create the limiter once where the listener is registered (in `enableCesiumOnTermination()`), pass it directly to `postRender.addEventListener`, and drop this wrapper method:

```ts
if (this.boundingSphere_) {
  const limiter = limitCameraToBoundingSphere(
    scene.camera,
    this.boundingSphere_,
    this.limitCameraToBoundingSphereRatio,
  );
  scene.postRender.addEventListener(limiter);
}
```

That preserves `blockLimiter` across frames as intended.

xytczh:
I agree with your analysis.Your proposed fix is right: create the limiter just once in enableCesiumOnTermination(), pass it directly to postRender.addEventListener, and drop the unnecessary wrapper method. As you pointed out, this will properly preserve the blockLimiter state across frames as intended.

ahocevar:
Thanks for updating, but this version is unfortunately worse: the `limitCameraToBoundingSphere()` wrapper method is itself already registered as the `postRender` listener (see the `addEventListener` call in `enableCesiumOnTermination()`). So on every frame it now creates a new limiter and adds it as *another* `postRender` listener — without ever calling it. The listener list grows by one every frame, and the constraint still never runs.

The wrapper method should be removed entirely. In `enableCesiumOnTermination()`, replace:

```ts
scene.postRender.addEventListener(
  this.limitCameraToBoundingSphere.bind(this),
);
```

with:

```ts
scene.postRender.addEventListener(
  limitCameraToBoundingSphere(
    scene.camera,
    this.boundingSphere_,
    this.limitCameraToBoundingSphereRatio,
  ),
);
```

The `boundingSphere_` null guard is already there (the whole block is inside `if (this.boundingSphere_)`). No wrapper method needed.

ahocevar (changes requested):
See comment above for requested changes.

